PR #2, opened 2017-08-08
Added option for BIOS from Altirra 2.90 as replacement for bootstrap OS

robmcmullen:
* not 100% compatible, but at least allows operation when real ROMs are not available

I'm sure this has been thought of before and I don't know if there's any politics involved between Altirra and atari800, but this was useful for me. This patch defaults to using the Altirra BIOS over the bootstrap OS.

The license from Altirra itself is GPL, although the license for his reimplemented ROMs is more permissive and compatible:
```
 *       Altirra - Atari 800/800XL emulator
 *       Kernel ROM replacement
 *       Copyright   (C)   2008-2016   Avery   Lee
 *
 *       Copying  and  distribution  of this file, with or without modification
 *       are permitted in any medium without royalty provided  the  copyright
 *       notice and this notice are preserved.  This file is offered as-is,
 *       without any  warranty.
```


knik0:
"Altirra BIOS - memo pad" looks OK to me.

joysfera:
Rob McMullen píše v Út 08. 08. 2017 v 19:21 +0000:
> not 100% compatible, but at least allows operation when real ROMs are
> not available

I've been thinking about this, too. Thank you for the patch/PR! This
will be a great addition to the new release.

Petr

robmcmullen:
Glad to contribute!

Issue #3, opened 2017-08-13
"Accurate skipped frames" setting is not persisted in configuration file


kcbrown:
When one elects to use a refresh rate other than 1:1, the "accurate skipped frames" option becomes available, to make it possible for collision detection to behave properly despite the lowered refresh rate.

That option is not persisted in the configuration file, though the refresh rate is.  As a result, when one wishes certain games that make use of collision detection to work properly, one must manually toggle "accurate skipped frames" upon starting the emulator.

The following patch fixes this:

```
diff --git a/src/cfg.c b/src/cfg.c
index bb07376..201e14a 100644
--- a/src/cfg.c
+++ b/src/cfg.c
@@ -176,6 +176,8 @@ int CFG_LoadConfig(const char *alternate_config_filename)
                                        Log_print("Unsafe PRINT_COMMAND ignored");
                        }

+                       else if (strcmp(string, "ACCURATE_SKIPPED_FRAMES") == 0)
+                               Atari800_collisions_in_skipped_frames = Util_sscanbool(ptr);
                        else if (strcmp(string, "SCREEN_REFRESH_RATIO") == 0)
                                Atari800_refresh_rate = Util_sscandec(ptr);
                        else if (strcmp(string, "DISABLE_BASIC") == 0)
@@ -387,6 +389,7 @@ int CFG_WriteConfig(void)

 #ifndef BASIC
        fprintf(fp, "SCREEN_REFRESH_RATIO=%d\n", Atari800_refresh_rate);
+       fprintf(fp, "ACCURATE_SKIPPED_FRAMES=%d\n", Atari800_collisions_in_skipped_frames);
 #endif

        fprintf(fp, "MACHINE_TYPE=Atari %s\n", machine_type_string[Atari800_machine_type]);
```

kcbrown:
Note that I've created a pull request to address this issue (it incorporates the previously mentioned patch): https://github.com/atari800/atari800/pull/4

PR #4, opened 2017-08-13
Previously, changes to Atari800_collisions_in_skipped_frames were not being saved

kcbrown:
Previously, changes to Atari800_collisions_in_skipped_frames were not being saved to the user's configuration file.  This has been fixed, and we now save to a variable called "ACCURATE_SKIPPED_FRAMES".


kcbrown:
This pull request corresponds to the issue here: https://github.com/atari800/atari800/issues/3

PR #9, opened 2017-09-20
Revitalised Atari Falcon port

mikrosk:
Please merge on the top of #8.

- Sound ThinAPI
- Videl fixes & RGB support
- SuperVidel support
- double/tripple buffering

I'm also volunteering to keep the Falcon port alive as a maintainer.


groessler:
Your code contains CR/LF line endings, e.g. src/falcon/c2p_uni.asm.

Please fix your git configuration and this PR.

mikrosk:
My git is fine. The file used to be present in the repository like this since the beginning, i.e. git's `autocrlf ` wouldn't catch this. One of the reasons is (was) that Devpac assembler can't work with LF line endings.

Now we're using vasm, so this limitation is gone.

groessler:
No, on the master branch src/falcon/c2p_uni.S has proper line endings.

No text file ever should be checked in with CR/LF. If you're running Windows, your git client should take care of that (make CR/LF endings for Windows when checking out).

Please check.

mikrosk:
c2p_uni.S and c2p_uni.asm are different files aimed for different compilers (with the latter predating the former and now removed again).

Anyway, it's fixed now.

groessler:
Thanks.

DOC/INSTALL.falcon is still CR/LFy...

mikrosk:
I've changed that as well but explicitly set to checkout as CR/LF as target audience is definitely CR/LF world (Atari TOS).

groessler:
Is there a git cleint for Atari TOS?

groessler:
If not, people will check it out on another system, and probably read it on that system.

groessler:
Another question: (This is my personal curiousity, and not meant to question the PR): What problems with GNU as did you encounter that you switched to vasm?

mikrosk:
git: sure it is: https://github.com/freemint/git (even if it weren't, who would remember to convert the documentation each release; remember it's supplied with the binary as well)

vasm: the sad fact is that we don't know precisely. @joysfera tried very hard in 2008: https://github.com/atari800/atari800/commit/67d2eb80a0611f562b6d8b9c14362822315278e8 but failed and eventually gave up, leading to 9 years of absence of Atari800 Falcon builds. I gave it a quick look as well but the resulting object file was just mess, really hard to see where the macros went wrong.

groessler:
git: If there is a client, you shouldn't need to force any file (INSTALL.falcon) to CR/LF. It should be handled by the git client.
Example: I (running Linux/Unix with LF line endings) want to change INSTALL.falcon for some reason. It should be a plain text file in the format my OS expects.

vasm: Ok. Since I sort of "like" GNU as (I'm a maintainer there) I could try to take a look as well.

mikrosk:
Well, how would you propose to handle it? FreeMiNT (Atari TOS ancestor) works in LF. So all the scripts and files are expected to be LF. Therefore git must clone files in LF. And yet when you make a release, majority of the users will use TOS only (which is CR/LF). Our platform has many specifics, it's something like a hybrid between MS DOS and Unix so you can't always apply rules from either world only.

If you could make the output from cpu_m68k.S 1:1 binary compatible to cpu_m68k.asm, that would be awesome. But be warned, it's trickier than it looks. ;-) If you need a quick setup, look here: vincent.riviere.free.fr/soft/m68k-atari-mint

groessler:
CR/LF: I admit, I didn't know about the differences btw, FreeMINT and TOS. I sort of know TOS. The solution is dependend on how you intend to make the "release" (which AFAIU should have CR/LF in its text files). Will you build/create it in an Atari itself, or a PC. And if PC, Windowish or Unixish?
(I've only got one 16bit Atari MegaSTE, since last year. I saved it from the dumpster. Guy ran along with the machine. I: "is it still working?". He: "yes". I: "can I have it?". He: "yes". I've played with it a little bit (installed gcc and bash), but since then it's sitting in a corner of my computer room,)

cpu_m68k.S: I'd like to take a look (not promising anything). Let's continue this part in personal mail...

mikrosk:
@groessler I guess that guy would have been very sad to realise that MegaSTEs go for ~300 EUR on ebay. ;)

groessler:
:-)  Didn't know that...

Do you think it would be possible to get atari800 working on the MegaSTE, or would this be a rather futile experiment?

mikrosk:
Define 'working' :). Certainly it would run but at what speed, that's another matter. The biggest problem is ironically not the lack of horse power (m68k is quite powerful even at 8 MHz) but the lack of a chunky oriented screen mode. With something like that I could imagine having a "hardware to hardware" emulator (i.e. reusing its VBL = VBI, HBL = DLI etc).

groessler:
With "chunky screen mode" you mean resolutions lower than 320x200? Which would mean less work for the CPU since the screen buffer is smaller. How is the Falcon different in this regard?

mikrosk:
I mean byte oriented screen. Amiga and Atari (except their last model, the Falcon) use bitplanes which make things much harder. Ironically, the Falcon port still uses them because its chunky mode is 16-bit ("hicolour"), i.e. no palette + bigger memory bandwidth.

joysfera:
IIRC the chunky2planar conversion has been optimized so much that it's
almost as efficient as plain memcpy.

Dne 28. 9. 2017 1:33 dop. napsal uživatel "Miro Kropacek" <
notifications@github.com>:

mikrosk:
@joysfera the point is that on the SuperVidel I don't do any c2p at all. Screen_atari serves as a video base pointer directly.

PR #21, opened 2018-01-27
GCC treats certain warnings as errors when instructed to do so.

KaleviKolttonen:
Fix printf() calls so that compiling works on Fedora Linux 27.


joysfera:
I happened to fix these printf() before I noticed your PR, sorry.
Thank you anyway!

PR #30, opened 2018-05-12
Allow SIO patch to be applied to Altirra OS

robmcmullen:
Added a check for Altirra OS in ESC_PatchOS to allow the fast SIO loading to work.


joysfera:
thanks!

PR #38, opened 2018-08-03
libatari800

robmcmullen:
This is my initial version of atari800 as a library. It creates a libatari800 configure target which generates a static library. It requires no external or platform-specific libraries to build.

It basically provides a way to generate successive video frames under the control of a user program. The user API is pretty thin at the moment, some minimal setup code, generating the next frame, and getting pointers to some of the internal products like the 64k main memory and the screen as stored in Screen_atari. It's up to the user program to do anything with the data.


joysfera:
I am curious what is the purpose of such library? Also, the patch has some conflicting files right now.

robmcmullen:
> I am curious what is the purpose of such library? 

Emulator can be built without any dependencies **or** used inside another program. I've been developing a cross-platform graphical debugger.

debugger

robmcmullen:
> Also, the patch has some conflicting files right now.

I hadn't been keeping this branch up to date because nobody had expressed interest in these past few months, but I'm happy to bring it up to date and maintain it if it is acceptable.

robmcmullen:
I also have a threaded debugger to get around the problem that a single threaded program can't handle both Atari800_Frame and PLATFORM_Exit.

joysfera:
cross platform debugger sounds great. Please update the conflicting files.

robmcmullen:
I'm working on getting this patch working with the new Automake configuration files. I might end up closing this patch and opening a new one just so it won't include the cruft of the old Autoconf stuff.

mikrosk:
You are free to force push ("overwrite") new history, within this PR. Just replace the branch with whatever changes you like, this PR will get updated accordingly.

pik33:
Libatari without OS dependencies would be good to port for a RPi bare metal system called Ultibo. I plan to do the port anyway, but having the platform independent code will speed up the process

robmcmullen:
> Libatari without OS dependencies would be good to port for a RPi bare metal system called Ultibo. I plan to do the port anyway, but having the platform independent code will speed up the process

Platform independent code was the goal. The only dependency this patch has is libm.

robmcmullen:
I just force-pushed over the old branch, so the patch is now updated to conform to the new Automake system. If compiled with `./configure --target=libatari800` it will create a static library and a simple test program that shows the embedding in action by emulating 200 frames as fast as it can.

I added a new README.libatari800 describing the changes and usage.

mikrosk:
I let @joysfera comment too but from my point of view -- wouldn't be more interesting to actually create the patch in that way that libatari800 would be used to link the actual executable of atari800? Therefore we wouldn't need to maintain yet another target and yet anyone could play with your library to his/her liking.

joysfera:
I haven't looked at the patch yet so I can't comment - I am just curious if such nice generic approach wouldn't add an unnecessary software layer that could slow the resulting binary down. Also, wouldn't it limit what @robmcmullen plans to do with the library? I mean in theory it sounds nice but too much abstraction could affect the performance.

mikrosk:
@robmcmullen will tell us about possible speed/abstraction trade-offs for sure. I'm familiar with such division in many open source projects, i.e. I take only the library part and ignore executable. Sometimes it is done via some configure switch, like `./configure --no-<project executable name>` or so.

robmcmullen:
The atari800 core is already pretty well abstracted, virtually everything machine-specific is in a subdirectory. I think of the libatari800 target as just another front-end. All libatari800 is really doing is providing an API to access the internals of atari800, so that external programs don't have to be compiled in with the atari800 source.

My particular use case for libatari800 is embedding atari800 in a Python program that provides the UI and debugger.

mikrosk:
I've briefly looked at the patch - I think you are right, Rob - it really makes sense to consider it as another target because that library isn't really a library. It reminds me another (existing) target, not sure how much supported these days - it was called "basic" or so? (IIRC, it's the `#ifdef BASIC` code) It was UI-less, exactly as your "library", rendered just frame. So maybe it would be worth reworking/extending that target?

robmcmullen:
Here's a view of what I'm using libatari800 for: embedding atari800 in a GUI to implement a graphical debugger. It uses save states stored in memory to implement the rewind feature. 
peek 2018-08-23 19-22

mikrosk:
> I am curious what is the purpose of such library? Also, the patch has some conflicting files right now.

Interestingly, there has been a pretty nice outcome also thanks to this library: https://github.com/pedgarcia/fujisan. Who would have thought. ;)

PR #47, opened 2018-10-13
Fixes typo in EMUOS_ALTIRRA defines

dmsc:
As subject says... this prevented applying the SIOV patch into Altirra OS roms.


mikrosk:
Nice catch, merged.

PR #50, opened 2018-10-16
debian rules update

hatchcliff:
I used dpkg-buildpackage to build a DEB file recently and had to edit debian/rules to get it to work.  I removed all instances of "cd src ; ".   I think these commands are no longer required because of the switch to out-of-source-tree compiling.  Configure and make are now in the atari800 directory, so need to change to src to access them.


hatchcliff:
Typo error in my previous comment, I meant:
"... so no need to change to src to access them."

Issue #46, opened 2018-10-09
Altirra BASIC broken


mikrosk:
- start without any configuration file
- during the boot press F1, Emulator configuration, Enable BASIC
- F5
- type "ABCD"
- type "BYE"
- F5
- type "DIR"

Atari800 crashes.

Built with:

`../atari800.git/configure --enable-veryslow --disable-monitorbreak --disable-monitorhints --disable-crashmenu --disable-monitorasm --with-readline=no --disable-eventrecording --disable-monitorutf8 --disable-monitoransi`

on a 64-bit Linux, fairly up to date libraries.

dmsc:
Hi!

I investigated this.

Problem is that a RESET from Altirra OS about screen does a warm-start instead of a cold-start.

mikrosk:
@dmsc is it something Altirra ROM should fix or is it Atari800's fault?

dmsc:
It is a problem in Altirra ROM, as it will also happen in real hardware. I will report to Phaeron.

dmsc:
Ok, reported here: http://atariage.com/forums/topic/281825-altirra-310-released/page-6#entry4132706

dmsc:
Hi!

Phaeron updated Altirra OS, fixing this and another issues, you can get the source/binary at http://atariage.com/forums/topic/281825-altirra-310-released/page-7#entry4144130

dmsc:
ping?

PR #51, opened 2018-11-04
Update Altirra OS to version 3.11 included in Altirra-3.20-test4.

dmsc:
Also, adds version information to Altirra BASIC and 5200 OS headers.

Fixes issue #46


dmsc:
ping?

PR #53, opened 2018-12-10
graphics9 bug on RPi patched

pik33:
When the Atari color in GTIA mode has values %1xxx1111, it was displayed as dark. This bug only occurred on the Raspberry Pi platform causing dark spots displayed where there should be light colors.
All colors >128 were shifted one color up, but while in not in  GTIA mode, this was not so easy to notice 
This patch causes all color displayed as they should be.


pik33:
I found these experimentally changing the pallette and observing the emulator. If the color is greater than 128, there is 1 position shift (displayed color=129). So I added 255/256 coefficient to compensate this skip.

What is added after the multiply is less important, it simply has to be >0 (and < 255/256? I didn't test this)

PR #61, opened 2019-06-18
Add POKEY recording functionality

mikrosk:
It supports mono and stereo recording of the first 9 Pokey registers to a file. This can be handy for recording (Turbo) Basic tunes or songs for which it is difficult to separate the player from the game.

Contributed by Ivo van Poorten.

First feature request here: https://atariage.com/forums/topic/242349-sap-type-r-recording-turbo-basic-tunes-to-sap

Published here: https://sourceforge.net/p/atari800/mailman/message/34458219/

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -75,6 +75,7 @@ dnl Silent rules
 m4_ifdef([AM_SILENT_RULES],[AM_SILENT_RULES([yes])])
 
 WANT_IDE="yes"
+WANT_POKEYREC="yes"
 
 dnl Set a8_host...
 
@@ -1060,6 +1061,15 @@ if [[ "$WANT_IDE" = "yes" ]]; then
 fi
 AM_CONDITIONAL([WANT_IDE], test "$WANT_IDE" = "yes")
 
+A8_OPTION(pokeyrec,$WANT_POKEYREC,
+          [Provide Pokey registers recording (default=ON)],
+          POKEYREC,[Define to add Pokey registers recording.]
+         )
+if [[ "$WANT_POKEYREC" == "yes" ]]; then
+    AC_SYS_LARGEFILE
+    OBJS="$OBJS pokeyrec.o"
+fi
+
 if [[ "$a8_use_sdl" = yes ]]; then
     A8_OPTION(onscreenkeyboard,no,
               [Enable on-screen keyboard (default=OFF)],
@@ -1267,6 +1277,7 @@ echo "Using event recording?................: $WANT_EVENT_RECORDING"
 echo "Using MIO emulation?..................: $WANT_PBI_MIO"
 echo "Using Black Box emulation?............: $WANT_PBI_BB"
 echo "Using IDE emulation?..................: $WANT_IDE"
+echo "Using Pokey registers recording?......: $WANT_POKEYREC"
 echo "Interface for sound...................: $with_sound"
 if [[ "$with_sound" != no ]]; then
     echo "    Using nonlinear mixing?...........: $WANT_NONLINEAR_MIXING"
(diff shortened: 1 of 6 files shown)


mikrosk:
Similar patch, I don't get what is the difference: https://sourceforge.net/p/atari800/patches/5 ?

mikrosk:
Argh, it fails because we moved from configure.ac `OBJS` to Makefile.am. @th-otto can you advise the easiest fix here?

th-otto:
Just add pokeyrec.[ch] to the list of source files:

```
diff --git a/src/Makefile.am b/src/Makefile.am
index 89e080f..a9177cd 100644
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -68,6 +68,7 @@ atari800_SOURCES = \
        pbi.c pbi.h \
        pia.c pia.h \
        pokey.c pokey.h \
+       pokeyrec.c pokeyrec.h \
        roms/altirra_5200_os.c roms/altirra_5200_os.h \
        rtime.c rtime.h \
        sio.c sio.h \
```
The location there is the list of source files that are always compiled, there are some others when using WITH_SOUND etc.

After that, just regenerate dependant files by running autogen.sh.

mikrosk:
Yeah but if we want to do it conditionally? (`WANT_POKEYREC` = yes)

th-otto:
Then just add a switch like AC_ARG_WITH([pokeyrec]) to configure.ac (just use some other switch there as example), add a AC_DEFINE(WITH_POKEYREC) that depends on that, and either use an AM_CONDITIONAL to conditionally add the files to list of source files, or just enclose the whole source file in #ifdef WITH_POKEYREC (after inclduing config.h of course)

mikrosk:
Gosh I hate that stuff. Many thanks, I'll fix it.

While I am at it, can you explain to me another autotools mystery? When I do `make dist`, it creates a nice release tarball. However two folders, `util` and `src/libatari800` leave `.deps` folder in there. Other folders normally also contain `.deps` folder but it is skipped, except those two.

th-otto:
For util, that is because of
```
EXTRA_DIST = $(doc_DATA) act data debian util
```
in the toplevel Makefile.am. That will just add everything in the util directory to the tarball. If you have actuallly sources there that are compiled, and thus produce tempory files, you should add a Makefile.am to that directory, add util to the SUBDIRS in the toplevel Makefile.am, add util/Makefile.am to the CONFIG_FILES in configure.ac, and then list any file that should be distributed explicitly in util/Makefile.am

When i run "make dist" here, i have only have that .deps directory in the tarball, but not the other.

th-otto:
BTW:
```
AC_INIT(Atari800, 4.0.0, pstehlik@sophics.cz)
```
The version number is still 4.0.0. That is the one used by automake do determine the name of the archive, amongst others.

joysfera:
@th-otto OK, will increase that version number upon next release.

th-otto:
Hm, that util/.deps directory seems to be some leftover of a different branch? I can just delete it, and run "make dist" in the master branch without problems.

mikrosk:
> Hm, that util/.deps directory seems to be some leftover of a different branch? I can just delete it, and run "make dist" in the master branch without problems.

It's possible you can delete it but it is certainly not a leftover, it is surely created by `./configure`.  However it must be somehow connected to the latest addition, `libatari800` (see https://github.com/atari800/atari800/pull/38) as `util/.deps` contains also `libatari800_test-libatari800_test.Po` file (and `libatari800` is the other `.deps` which made it into the archive).

mikrosk:
@th-otto faster than me, thanks!

PR #63, opened 2019-06-18
libatari800 fixes

robmcmullen:
Fixes for a compiler warning and for a new UI_ variable added in a recent commit


joysfera:
thanks!

PR #64, opened 2019-06-18
Added documentation for demo programs included with libatari800

robmcmullen:
Added more detail about the `guess_settings` demo program that uses libatari800 to run through a bunch of permutations of emulator settings to attempt to guess which combination of settings will successfully boot the image.


mikrosk:
Simple enough, merged.

PR #67, opened 2019-06-20
Make dist fix

th-otto:


th-otto:
Argl mom, something wrong here.

joysfera:
Now this is an interesting PR...

PR #72, opened 2019-08-25
use zlib-provided crc32 if HAVE_LIBZ is defined

asiekierka:
This has the potential to save some code/data space (up to ~1K) in the final binary, and avoid redundant implementations of the same algorithm.

~~EDIT: Doesn't seem to work quite right - I wonder why...~~

EDIT 2 (05/06/2020): Seems to work fine! However, I cannot test every case.

### DIFF
--- a/src/crc32.c
+++ b/src/crc32.c
@@ -42,6 +42,8 @@
 		table[i] = crc;
 	}
  */
+
+#ifndef HAVE_LIBZ
 static ULONG table[0x100]= {
 	0x00000000, 0x77073096, 0xee0e612c, 0x990951ba, 0x076dc419, 0x706af48f, 0xe963a535, 0x9e6495a3,
 	0x0edb8832, 0x79dcb8a4, 0xe0d5e91e, 0x97d2d988, 0x09b64c2b, 0x7eb17cbd, 0xe7b82d07, 0x90bf1d91,
@@ -86,6 +88,7 @@ ULONG CRC32_Update(ULONG crc, UBYTE const *buf, unsigned int len)
 
 	return crc;
 }
+#endif
 
 int CRC32_FromFile(FILE *f, ULONG *result)
 {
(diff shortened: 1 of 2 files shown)


mikrosk:
> EDIT: Doesn't seem to work quite right - I wonder why

Well, this makes this PR then quite useless, doesn't it. ;) Please fix it or close it.

mikrosk:
@asiekierka any chance to look at this?

asiekierka:
Apologies, only had time to look at it now.

It's been a while. I *think* (and a brief test seems to confirm that) the [second commit](https://github.com/atari800/atari800/pull/72/commits/0a86499da55cf40038609fa19dbe981fa76238e1) fixed the issue I mentioned in the EDIT; just that I forgot to edit the PR back. (The way to test this, IIRC, is to see if some ROMs get accepted by the auto-detection).

Whether this PR is desirable to the project is another story - I looked into it for my console ports, where a small size saving and not duplicating code can make sense, but your mileage may vary.

dmsc:
Hi!
> Apologies, only had time to look at it now.
> 
> It's been a while. I _think_ (and a brief test seems to confirm that) the [second commit](https://github.com/atari800/atari800/pull/72/commits/0a86499da55cf40038609fa19dbe981fa76238e1) fixed the issue I mentioned in the EDIT; just that I forgot to edit the PR back. (The way to test this, IIRC, is to see if some ROMs get accepted by the auto-detection).
> 
> Whether this PR is desirable to the project is another story - I looked into it for my console ports, where a small size saving and not duplicating code can make sense, but your mileage may vary.

IMHO, this is desirable.

You should squash both commits and update the PR description.

Thanks!

mikrosk:
No need to squash it, I'll do it when merging. So are you saying it has been tested and it is ready to merge?

asiekierka:
To the extent I could test it, yes. It seems it's only used for ROM auto-detection, so that's where I verified it. I can't guarantee it works with every ROM or case, though.

mikrosk:
In that case, let's merge it!

Issue #29, opened 2018-04-23
Missing command line options.


severach:
Years ago I was unable to get Linux Atari 800 emulators working. The reason is that the GUI fine for testing and small changes but doesn't work well for the extensive initial setup. The command line exposes more settings which always work once set up. Unfortunately I found that a few settings don't have command line options. Here is the sed line I used to fill in the missing settings. There may be more but these are the ones that I wanted.

````
  # No command line switches for some settings
  sed -e 's:^\(SDL_JOY_0_ENABLED\)=.*$:\1=0:g' \
      -e 's:^\(ATARI_FILES_DIR\)=.*$:\1=/home/chris/c/games/atari/romsdave:g' \
      -e 's:^\(SAVED_FILES_DIR\)=.*$:\1=/home/chris/c/games/atari:g' \
    -i ~/'.atari800.cfg'
````

joysfera:
maybe you could explain your issue and fix of it in plain words. The `sed` command just disables first joystick for me. I doubt that an enabled joysticks would not allow you to run Atari 800 for years.

severach:
The original issue is that it is nearly impossible to set up the emulator using the on screen configuration. After about an hour of config and testing, press the wrong button, and it all disappears. When you want to copy the config to another machine all you have is outdated junk.

This is irrelevant because the command line is the right way to set up the emulator. It still takes an hour to get all the settings right but it doesn't all disappear when you press the wrong button and once done you can copy it anywhere, adjusting a few settings to match the computer. Not only that, many of the needed settings are not found anywhere in the config UI. They are only available from the command line.

The current issue is that these settings do not have command line options. The only way I can set them is to start the emulator which writes `.atari800.cfg` then modify it then start the emulator again.

ChoccyHobNob:
Which settings are you talking about? There are only a couple that cannot be set from the command line (1088K ram, stuff like that) which ones can only be set from the command line?

The most important setting for you is probably the /config setting which allows you to specify a config file to use

severach:
Settings found in the GUI but missing from the command line:
````
SDL_JOY_0_ENABLED, default 1, 0 to disable joystick keyboard emulation
ATARI_FILES_DIR, the folder that Alt-R starts with.
SAVED_FILES_DIR, the folder that load and save start with.
````
SDL_JOY_0_ENABLED does not disable the joystick unless you have a keyboard based joystick.

Settings not in the GUI, only accessible from the command line:
````
-win-width
-win-height
-fs-width
[...]

PR #77, opened 2019-12-27
Add missing command line options

mikrosk:
As per issue #29, this patch adds following command line options:

- `-atari_files` (analog to `ATARI_FILES_DIR`)
- `-saved_files` (analog to `SAVED_FILES_DIR`)
- `-kbdjoy0` (analog to `SDL_JOY_0_ENABLED = 1`)
- `-kbdjoy1` (analog to `SDL_JOY_1_ENABLED = 1`)
- `-no-kbdjoy0` (analog to `SDL_JOY_0_ENABLED = 0`)
- `-no-kbdjoy1` (analog to `SDL_JOY_1_ENABLED = 0`)

@severach can you confirm this is what you had in mind? If so, we can merge this PR and consider #29 as fixed. The source tree to compile can be found at https://github.com/atari800/atari800/tree/issue29.

### DIFF
--- a/src/atari.c
+++ b/src/atari.c
@@ -770,6 +770,7 @@ int Atari800_Initialise(int *argc, char *argv[])
 #endif
 #if !defined(BASIC) && !defined(CURSES_BASIC)
 		|| !Screen_Initialise(argc, argv)
+		|| !UI_Initialise(argc, argv)
 #endif
 		/* Initialise Custom Chips */
 		|| !ANTIC_Initialise(argc, argv)
--- a/src/sdl/input.c
+++ b/src/sdl/input.c
@@ -1283,6 +1283,18 @@ int SDL_INPUT_Initialise(int *argc, char *argv[])
 			else a_m = TRUE;
 		}
 #endif /* LPTJOY */
+		else if (strcmp(argv[i], "-kbdjoy0") == 0) {
+			PLATFORM_kbd_joy_0_enabled = TRUE;
+		}
+		else if (!strcmp(argv[i], "-kbdjoy1")) {
+			PLATFORM_kbd_joy_1_enabled = TRUE;
+		}
+		else if (strcmp(argv[i], "-no-kbdjoy0") == 0) {
+			PLATFORM_kbd_joy_0_enabled = FALSE;
+		}
+		else if (!strcmp(argv[i], "-no-kbdjoy1")) {
+			PLATFORM_kbd_joy_1_enabled = FALSE;
+		}
 		else {
 			if (strcmp(argv[i], "-help") == 0) {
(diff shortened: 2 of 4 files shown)


severach:
Looks pretty good. I see a bug.

Each run of `atari800` works fine but I never run it that way. I always lauch atari800 with a full set of switches. Each run of `atari800 -atari_files` or `atari800 -saved_files`  adds a duplicate entry into the config file. After UI_MAX_DIRECTORIES (8) runs this results in

````
All ATARI_FILES_DIR slots used!
All SAVED_FILES_DIR slots used!
````

mikrosk:
Hmm, this indeed looks like a bug. I assume you would like to see the adding to the config file skipped, correct?

I still wonder how this is possible. I could imagine that the first time the config file is created, perhaps the switches have been parsed before that (and their content stored) but why the config file is changed just by the values on the command line, that's a mystery. I guess there is an immediate write command somewhere in the directory handling code what looks wrong, the config file should be updated only when explicitely asked for.

severach:
According to the rules I should be able to specify `-saved_files` up to 8 times and they should all go in. More than 8 should generate warnings. dups shouldn't be added.

I looked at the new code and I can see that the command line folder is just added. There's no check through the existing ones to see if the folder is already present.

mikrosk:
> Each run of atari800 -atari_files or atari800 -saved_files adds a duplicate entry into the config file.

I'm sorry but I cannot reproduce this, nothing has been saved into the config files, as expected.

It's true there is no check for duplicate entries but I fail to see why somebody would want to intentionally put more than 8 identical directory names on the command line and complain that they have been rejected.

Solution is simple, don't put identical or too many names on the command line. :) If you create a PR, I guess it can be merged but I don't have motivation to fix this, it's really a corner case.

severach:
I'm not getting dups on the same run. I'm getting them on successive runs.

````
rm ~/.atari800.cfg
atari800 -autosave-config -atari_files foo
atari800 -autosave-config -atari_files foo
grep ATARI_FILES_DIR ~/.atari800.cfg
````

mikrosk:
My question still stands: why would you do that? You claimed that you want those command line parameters because you don't use the config file and now you clearly do so now I'm really lost what is the use case here. ;-)

severach:
Here's the patch. I fixed another bug when "all slots used" the path is interpreted as a parameter.

[0000-save.files.no.dups.patch.zip](https://github.com/atari800/atari800/files/4052220/0000-save.files.no.dups.patch.zip)

mikrosk:
Thanks - any chance to make a new PR from the patch? So it wouldn't get lost (this one is pretty much closed).

PR #86, opened 2020-01-13
replace caddr_t with standard type

Redfoxymoon:
caddr_t is a legacy BSD type, there's no point in keeping it when char* is identical


mikrosk:
What's wrong with casting it to `struct in_addr*` ?

Redfoxymoon:
I am not sure I understand? There's no functional change, it's only a portability change.

mikrosk:
Oops, sorry, I have confused myself. What I meant why the casting at all? `memcpy` is perfectly happy with accepting any pointer (it's `void*` after all) so why not just remove the cast and `caddr_t`.

As it is faster to do than to talk about it, I just did it. :)

Issue #89, opened 2020-02-02
NTSC Filter Presets cause screen to be washed out


fragmental:
With the ntsc filter on, selecting any of the presets will cause the screen to look washed out, or white.  
Example:
![Example](https://i.imgur.com/cQeGr24.png)

In order to work around this, I can then select `Restore default colors` and edit the filter values manually.  I have tested this with version 4.2.0 and also the current retroarch core.

dmsc:
Hi!

Yes, when you select a NTSC filter preset, the GAMMA is set to 0.3, this causes the screen to look washed out.

Will investigate.

dmsc:
The bug was introduced at revision 5f6d891ac1023917133c2ec560dad3eaa27e90d7 , that changed gamma processing but did not update the NTSC filter presets.

PR #90, opened 2020-02-02
Update gamma values in NTSC filter presets to the new values.

dmsc:
This fixes issue #89, introduced in commit 5f6d891ac1023917133c2ec560dad3eaa27e90d7


mikrosk:
Looks good, thanks.

PR #93, opened 2020-03-08
Cartridge ROM to CART converter

wfd3:
I've created a little tool to convert raw cartridge ROMs to the CART file format.  I've added this to a new directory .../tools, and provided an Automake stub to integrate into the build system.


wfd3:
Missed a file.  Will fix and resubmit.

PR #100, opened 2020-05-03
Added tool to create carfiles from romfiles

wfd3:
This change adds tools/cart.c, a utility to convert romfiles to cartfiles.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,7 @@
 ACLOCAL_AMFLAGS = -I m4
 AUTOMAKE_OPTIONS = foreign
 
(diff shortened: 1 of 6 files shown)


mikrosk:
Oh, you seem to create the same problem again :-) If you want to create two separate PRs, you have to create two separate branches. :)

So in your case - switch to your old master, create a branch, cherry-pick your first change, then switch to your old master again, create another branch and cherry-pick you second change.

wfd3:
Apologies.  Obviously new to Git.  Will take another pass at this, hopefully will get it right this time.

PR #101, opened 2020-05-03
Add tool to create cart files from romfiles

wfd3:


tools/cart.c
```diff
@@ -0,0 +1,139 @@
...
+  uint8_t *rom;
+  uint32_t csum;
+  header h;
+
+  printf("Converting %s to %s (type %d)\n", romfile, cartfile, type);
+
+  /* Read the original rom file */
+  fp = fopen(romfile, "rb");
+  if (fp == NULL) {
+    printf("Error opening rom file '%s': %s\n", romfile, strerror(errno));
```
dmsc, line 65: IMHO, you should use `fprintf(stderr,` for the error messages.

tools/cart.c
```diff
@@ -0,0 +1,139 @@
...
+    exit(1);
+  }
+  len = flen(fp);
+  rom = malloc(len);
+  if (rom == NULL) {
+      printf("Allocation error: %s\n", strerror(errno));
```
dmsc, line 71: Same here and in the rest of errors.

tools/cart.c
```diff
@@ -0,0 +1,139 @@
...
+
+  printf("Converting %s to %s (type %d)\n", romfile, cartfile, type);
+
+  /* Read the original rom file */
+  fp = fopen(romfile, "rb");
+  if (fp == NULL) {
+    printf("Error opening rom file '%s': %s\n", romfile, strerror(errno));
+    exit(1);
+  }
+  len = flen(fp);
```
dmsc, line 68: You should check if the length is < 2048 or > 0x8000000 and show an error. Or better yet if you check that the size is compatible with the cartridge type given.

wfd3: Done.  Please take another look?

mikrosk:
@dmsc is right, it would be shame to miss the opportunity to make your tool flexible enough to work with all cart types supported by atari800, now or in the future.

So if I can yet test your patience, please follow @dmsc's advice. :)

wfd3:
I'm still testing these changes but I wanted to give you all an early chance to offer feedback before I'm done.  I'm not sure how you would like this to look stylistically; happy to make whatever changes you'd like.

mikrosk:
Very nice. Just make sure to thoroughly test the UI code changes as those would be most visible to the user.

wfd3:
Alright, these are the final changes.  I've tested the UI as requested and it appears to be working.

src/cartridge_info.c
```diff
@@ -0,0 +1,111 @@
+/*
+ * cartridge_info.c - cartridge information
+ *
+ * Copyright (C) 2001-2010 Piotr Fusik
+ * Copyright (C) 2001-2010 Atari800 development team (see DOC/CREDITS)
```
dmsc, line 5: Minor nit: you could use 2020 as the last year, as you are creating a new file.

src/cartridge_info.h
```diff
@@ -0,0 +1,119 @@
+/*
+ * cartridge_info.h - cartridge information
+ *
+ * Copyright (C) 2001-2010 Piotr Fusik
+ * Copyright (C) 2001-2010 Atari800 development team (see DOC/CREDITS)
```
dmsc, line 5: Same here for the year, 2001-2020.

src/ui.c
```diff
@@ -1007,94 +1007,34 @@ static void DiskManagement(void)
...
-			menu_array[i - 1].flags = UI_ITEM_ACTION;
+				option = cart_entry;
+			flag = UI_ITEM_ACTION;
 		}
 		else
-			menu_array[i - 1].flags = UI_ITEM_HIDDEN;
+			flag = UI_ITEM_HIDDEN;
 
+		UI_MAKE_MENU_ACTION(menu_array[menu_entry], flag, cart_entry,
+				    CARTRIDGES[cart_entry].description);
```
dmsc, line 1033: This file is formatted with tabs of width 4 (see the last line in the file), so this does not look aligned.

src/ui.c
```diff
@@ -1007,94 +1007,34 @@ static void DiskManagement(void)
...
+		if (CARTRIDGES[cart_entry].kb == k) {
 			if (option == 0)
-				option = i;
-			menu_array[i - 1].flags = UI_ITEM_ACTION;
+				option = cart_entry;
+			flag = UI_ITEM_ACTION;
 		}
 		else
-			menu_array[i - 1].flags = UI_ITEM_HIDDEN;
+			flag = UI_ITEM_HIDDEN;
```
dmsc, line 1030: As you are dynamically creating the menu_array, you could simply include only the cartridge types actually supported, and skip the other ones, something like this:
```
	for (i = 1, menu_entry = 0; i <= CARTRIDGE_LAST_SUPPORTED; i++)
		if (CARTRIDGES[i].kb == k) {
			UI_MAKE_MENU_ACTION(menu_array[menu_entry], UI_ITEM_ACTION, i, CARTRIDGES[i].description);
			menu_entry++;
		}
	if(menu_entry == 0)
		return CARTRIDGE_NONE;
	UI_MAKE_END(menu_array[menu_entry]);
```

src/ui.h
```diff
@@ -146,6 +146,20 @@ typedef struct
 #define UI_MENU_SUBMENU_SUFFIX(retval, item, suffix)           { UI_ITEM_SUBMENU, retval, NULL, item, suffix }
 #define UI_MENU_SUBMENU_ACCEL(retval, item, keystroke)         { UI_ITEM_SUBMENU, retval, NULL, item, UI_MENU_ACCEL(keystroke) }
 #define UI_MENU_END                                            { UI_ITEM_END, 0, NULL, NULL, NULL }
+#define UI_MAKE_MENU_ACTION(a, f, r, i) {	\
+    a.flags = f;				\
+    a.retval = r;				\
+    a.prefix = NULL;				\
+    a.item = i;					\
+    a.suffix = NULL;				\
+  }
```
dmsc, line 155: Are you sure it is not enough to simply use the above macros and do: `my_item = UI_MENU_ACTION(ret, item);` The same for UI_MAKE_END.

src/ui.h
```diff
@@ -146,6 +146,20 @@ typedef struct
...
+#define UI_MAKE_END(a) {			\
+    a.flags = UI_ITEM_END;			\
+    a.retval = 0;				\
+    a.prefix = NULL;				\
+    a.item = NULL;				\
+    a.suffix = NULL;				\
+  }
```
dmsc, line 162: Are you sure it is not enough to simply use the above macros and do: `my_item = UI_MENU_ACTION(ret, item);` The same for `UI_MAKE_END`.

src/ui.h
wfd3: Yeah, I am.  Composite literals like you describe were introduced in ISO C99; the default behavior in the Makefile is to use -ansi, which is equivalent to -std=c90.   I've built and tested with -std=c18, the latest ISO standard, and  with -std=gnu18, which enables the GCC extensions to C18.  

I can verify it works fine on Linux, but I cannot test on other platforms. 

My view is that changing the language standard is probably safe but beyond what I can verify and that it's best to either use the implementation here or just do the assignment to members of UI_tMenuItem directly.  

Let me know how you want me to proceed. I'll make the other changes as suggested.

src/ui.c
wfd3: Sorry, are we talking about tab-stops in my deltas?  Or are you looking at the way the line break in UI_MAKE_MENU_ACTION() is aligned?

src/ui.h
wfd3: Ah, I forgot to mention that if you do move to -std=c18, the same composite literal will work for both structure assignment and initialization.

src/ui.c
dmsc: Yes, in my editor, the "CARTRIDGES" line is not aligned with "menu_" above, as it uses tab width = 4, because at the end of the file it says "vim:ts=4:sw=4:"

Sadly, the Atari800 sources use a mixture of tab and spaces of different widths all over the place.

src/ui.h
dmsc: I suspected that was the reason. I don't know if supporting old C90 is really needed, but perhaps one of the older ports (using old microsoft C compilers) do not understand composite literals.  @mikrosk what do yo think?

src/ui.c
wfd3: OK, that's just how my emacs is setup. It's an easy fix. :)

src/ui.h
wfd3: One thing to consider, albeit probably minor.  If you were to move to composite literals for the menu initializers, you'd have drop the static & const keywords from all declarations of UI_tMenuItem (70-something instances).

dmsc: My idea is somewhat simpler, just in this place do:
```
    for (i = 1, menu_entry = 0; i <= CARTRIDGE_LAST_SUPPORTED; i++) {
        if (CARTRIDGES[i].kb == k) {
            menu_array[menu_entry] = (UI_tMenuItem)UI_MENU_ACTION(i, CARTRIDGES[i].description);
            menu_entry++;
        }
    }
    menu_array[menu_entry] = (UI_tMenuItem)UI_MENU_END;
```
I tried a few other options, sadly all produced the some warning in C90 mode.
IMHO, this is the less ugly one:
```
	UI_tMenuItem menu_array[CARTRIDGE_LAST_SUPPORTED+1];
	UI_tMenuItem menu_end = UI_MENU_END;
	int i, menu_entry;
	int option = 0;

	UI_driver->fInit();

	for (i = 1, menu_entry = 0; i <= CARTRIDGE_LAST_SUPPORTED; i++) {
		if (CARTRIDGES[i].kb == k) {
			UI_tMenuItem menu_act = UI_MENU_ACTION(0, NULL);
			menu_act.retval = i;
			menu_act.item = CARTRIDGES[i].description;
			menu_array[menu_entry] = menu_act;
			menu_entry++;
		}
	}
	/* Terminate menu_array */
	menu_array[menu_entry] = menu_end;
```

wfd3: Yeah, I'm thinking along the same lines as your second example (although I'd move the menu_act initialization outside the loop).  It's more verbose but much easier to understand.

wfd3:
OK, that should cover all the open comments/suggests to date.  Let me know if you see anything else...

src/cartridge_info.c
mikrosk: This looks like being forgotten to update.

src/cartridge_info.h
mikrosk: Ditto

mikrosk:
Looks great to me, unless @dmsc has any other notes, I'll merge it today.

dmsc:
Yes, looks great!
Thanks @wfd3 for all your work!

mikrosk:
Thank you @wfd3 for your contribution & patience and @dmsc for checking the code.

wfd3:
No problem, happy to help.  Thanks for the feedback.

PR #102, opened 2020-05-03
Adds support for remapping OPTION, SELECT, START, RESET and EXIT keys, and adds comments to config file parser.

wfd3:


src/cfg.c
```diff
@@ -126,6 +126,10 @@ int CFG_LoadConfig(const char *alternate_config_filename)
 	while (fgets(string, sizeof(string), fp)) {
 		char *ptr;
 		Util_chomp(string);
+		/* Check for comments */
+		if (string != NULL && string[0] == '#') {
+			continue;
```
dmsc, line 131: This only detects '#' on the first column, it is as intended?
If not, you could add `Util_trim(string);` just before.

wfd3: Fixed.  Please take another look when you can.

mikrosk:
Thank you both!

PR #103, opened 2020-05-06
Small fix in saving config

dbroda:
After compiling in Fedora 31 (SDL), threre was an issue with saving config and using SELECT key.


mikrosk:
Thanks, appreciated.

Issue #106, opened 2020-05-21
Support for Atari 5200 bank switchable carts. 


TheRobotFactory:
Back in 2005 a bank switching cartridge was created for the Atari 5200 allowing up to 512KB. There has been some uptake in the Homebrew community but currently Real hardware or Altirra are the only way to run these cart images.

Details of the board are here https://atariage.com/forums/topic/69649-5200-game-announcement-now-shipping/?do=findComment&comment=856153

Example carts are attached. 

[Examples.zip](https://github.com/atari800/atari800/files/4660654/Examples.zip)

joysfera:
Should be implemented now thanks to @asiekierka

PR #108, opened 2020-06-05
Add support for 64-512K Atari 5200 bank-switchable carts (#106)

asiekierka:
It appears that state load/save and the examples linked in the issue work. Atari Falcon CPU core patches would need to be applied separately (similarly to Bounty Bob), as I do not have enough confidence in m68k assembly nor do I have a Falcon testing setup.

Paging @TheRobotFactory - as the Atarimax Ultimate SD (second-hand information) and your example set only limits itself to the 32K bankswitching method (and, as such, 64/128/256/512K ROMs), is implementing the 16K bankswitching variant required for any released game?


TheRobotFactory:
This is every bank switched cart image I have found. I don't know if any of them use the 16K variant or not. 
[BankSwitched5200.zip](https://github.com/atari800/atari800/files/4733295/BankSwitched5200.zip)

There is another type specifically for the UltimateSD that runs a cart image and loads data from another file, there's an example of that https://atariage.com/forums/topic/237766-bad-apple-ultimate-sd , the homebrew release of Atari Blast does the same thing, I don't expect this is in scope of the change as it would require some kind of emulation of the ultimate SD

asiekierka:
I think we could use someone with more experience pitching in. I think it's gonna be fine like this, as this increases support regardless.

TheRobotFactory:
I agree. 

Maybe change the name of the cart types to say that they are the 32K bank type not the 16K bank

src/cartridge_info.h
```diff
@@ -28,7 +28,7 @@
 /* Min and Max cartridge size in bytes */
 #define CARTRIDGE_MIN_SIZE        2048
 #define CARTRIDGE_MAX_SIZE	  (128 * 1024 * 1024)
-#define CARTRIDGE_LAST_SUPPORTED  70
+#define CARTRIDGE_LAST_SUPPORTED  74
```
mikrosk, line 31: I wonder why this isn't part of the enum?

asiekierka: I don't know... Could probably be refactored - should I do that?

mikrosk: Sure, code improvements are always good. :)

TheRobotFactory:
I found this today too, details of the Atarimax Ultimate SD 8k bank switching mode and examples https://atariage.com/forums/topic/221479-5200-ultimate-sd-cart-hybrid-8k-write8k-read-mode/?tab=comments#comment-2917347

src/cartridge_info.h
asiekierka: Done - as a separate commit so you can see my proposed changes.

asiekierka:
Pushed refactoring and documentation.

Regarding the Atarimax Ultimate SD - the fact it allows full writes to the cartridge's SRAM may necessiate a small redesign of how memory state for cartridges is deserialized; I would open a separate issue for it, myself.

TheRobotFactory:
There is another cart type for Atari 8-bit that allows writes to sram, called corina. Details are here, https://atariage.com/forums/topic/135561-corina-new-cartridge-architecture
I've only ever seen two carts using this type, "Bomb Jake" and "Ye Another Kung-Fu". There is apparently a special version of Atari++ that supports them and Altirra has support for them too. 

For the Ultimate SD the only titles currently using 'hyb' mode are Bad Apple, Atari Blast! and Bug Hunt. Bad Apple and Atariblast have a separate datafile that has to go on the SD card which gets loaded by the .hyb file. Bug hunt uses hybrid mode but doesn't load an external dat file.

The Corina carts are pretty old ~2008 and it doesn't look like the format took off GR8 software don't appear to be around any more so I doubt there will be any more Corina carts to support. The Ultimate SD is still pretty new, the hyb format titles were in 2015/16 so fairly recent, I'd expect that there will be more of those to come from homebrew developers in the future.

If there is any will to add Corina and the Ultimate SD formats I'll create an issue for it.

asiekierka:
An issue is always worth making. In general, adding cartridge write support isn't necessary just for Corina and Ultimate SD - the existing Atarimax and SIC! cartridge emulation is missing write support as well.

asiekierka:
Are there any plans to merge this, or is it waiting on something else?

joysfera (approved):
I'm a little bit concerned about the CARTRIDGE_TYPE_COUNT (must not be forgotten to be increased next time a new cartridge type is added) but otherwise it looks good to me.

asiekierka:
This should be easier to maintain still, as now it's part of the very same `enum` - earlier, CARTRIDGE_LAST_SUPPORTED was a separate `#define`.

Thank you!

PR #113, opened 2020-07-01
Update Altirra 800 and XL Kernels to the current version.

dmsc:
Adds a new folder with the full Altirra Kernel sources, and a makefile to assemble the sources and generate the C files in src/roms

Using this, import all OS revisions up to the last version released, fixing many bugs when using the internal roms.


mikrosk:
Doesn't Altirra use git by any chance? If so, we could make it a submodule.

dmsc:
Hi!

> Doesn't Altirra use git by any chance? If so, we could make it a submodule.

Not as far as I know, Avery publishes new beta versions over the AtariAge forum and makes the releases available in the web page.

If you like his, I can do the same for the Altirra BASIC rom.

Have Fun!

asiekierka:
I think it's a very good idea to make the ROMs buildable from source as part of the atari800 tree. It will also make it easier to see differences in updates.

mikrosk:
As I'm not a huge fan of that Altirra OS I guess I'm a bit biased here. But sure, why not. My only worry is that it adds another dependency and that is `mads`, correct?

dmsc:
Hi!

> As I'm not a huge fan of that Altirra OS I guess I'm a bit biased here. But sure, why not. My only worry is that it adds another dependency and that is `mads`, correct?

Yes, but it is only needed if you want to rebuild the ROM, it is not needed for the standard build, and never automatically done.

Have Fun!

dmsc:
Hi!, any update on this?

mikrosk:
Since nobody is against, I'm giving it green light.

PR #118, opened 2020-07-20
video accel default

hatchcliff:
This pull request is to fix a problem with installation on the Raspberry Pi 4.  Video acceleration is turned off by default, but the Pi 4 is not fast enough to run the emulator properly without it.  This can be put right by turning acceleration on later by hand, but it would be better if this was done automatically.

It appears that the simplest solution is to reverse the default, such that acceleration is turned on for all machines that support it, including the Pi 4.  The proposed change causes 'VIDEO_ACCEL=1' to be written into the configuration file when it is first created, instead of 'VIDEO_ACCEL=0'.  It has no other effects.  The setting can still be changed manually if required, and the options -video-accel and -no-video-accel still override the setting in the configuration file.

### DIFF
--- a/src/sdl/video.c
+++ b/src/sdl/video.c
@@ -68,7 +68,7 @@ static VIDEOMODE_resolution_t desktop_resolution;
 
 #if HAVE_OPENGL
 int SDL_VIDEO_opengl_available;
-int SDL_VIDEO_opengl = FALSE;
+int SDL_VIDEO_opengl = TRUE;
 /* Was OpenGL active previously? */
 static int currently_opengl = FALSE;
 #endif


mikrosk:
I maintain a platform which doesn't use acceleration at all so I don't mind. However one could argue that if someone has broken drivers (think various Linux/BSD environments), the emulator would start corrupted and maybe even not that (core dumped). So from this point of view it looks to me be "better safe than sorry". But I'm open to arguments in favour of enabling it.

joysfera:
The RPi is basically a single-OS platform (with the Raspberry Pi OS) so it if works there with acceleration enabled then it can be enabled by default, IMHO.

hatchcliff:
If turning on video acceleration were to cause some systems to crash, that would be a worse problem than the one I'm trying to fix - so I think I will have to do this differently.

I could make the change specific to the RPi4 and its single OS, but maybe there are other systems (current or future) that would benefit from the same treatment.

Do you think a general solution along the following lines might work?

Change configure.ac to recognise a new option, --accel-default-on.

In the part of configure.ac that defines HAVE_OPENGL (to indicate that hardware acceleration is available) nest some additional code to define ACCEL_DEFAULT_ON if the above option has been specified.

Modify video.c to default 'video acceleration on' in the newly created configuration file if HAVE_OPENGL and ACCEL_DEFAULT_ON are defined.

Edit BUILD.RPI to advise use of the --accel-default-on option for RPi4s, and the same for any other platforms that would benefit.

mikrosk:
Maybe it is just me being too paranoid. If @joysfera doesn't mind enabling it on all platforms, I can't really offer any data supporting my paranoia. :)

joysfera:
I was talking about RPi4 only. The original poster was also mentioning RPi4 only. The patch, if any, should address RPi4 only, nothing else.

mikrosk:
The change is in `src/sdl/video.c` for all OpenGL-aware targets.

hatchcliff:
Sorry, I seem to have created some confusion with my first post.  I started out looking for the simplest possible fix, which appears to be to turn on acceleration for all OpenGL-aware targets, but I accept this is not the best idea and we need to focus on the RPi4 only.

The current default is to turn off acceleration for all targets.  This is set by the declaration 'int SDL_VIDEO_opengl = FALSE;' at the start of src/sdl/video.c.  I think I need to change this to reverse the default for the RPi4, which probably means going back to configure.ac to change the way the conditional compilation of this and related declarations is controlled.  I'm not sure yet whether to manage this with a new option like --accel-default-on (as mentioned previously), or to create a new target option --target=rpi4 and handle it that way.  I will start experimenting with some code and try to work out the neatest solution.

asiekierka:
What I would recommend doing is enabling OpenGL by default (if present on a given platform in a high enough version), and keeping a list of OpenGL renderers which are known to be broken. [A project called MegaZeux uses this](https://github.com/AliceLR/megazeux/blob/master/src/render_glsl.c#L120-L157) uses such a list, as the "GLSL" renderer is the default, to fallback to the software renderer on specific platforms.

hatchcliff:
Thanks for your recommendation and link, which seem to confirm that the reservations expressed about broken OpenGL drivers are correct - not just paranoia.

I will shortly post a modification that turns acceleration off by default (as before) but allows us to make exceptions in the case of the RPi4 and any other machines with the same performance limitation.  This is not as thorough as your approach.  It will still permit the user to turn on acceleration on systems with broken drivers, but it avoids the more serious problem of crashes occurring immediately after installation with no clue as to the cause.  On the plus side, it does not require us to maintain a list of good/bad drivers.  I think it's probably worth consideration on grounds of simplicity - something to discuss further.

hatchcliff:
I've posted some new code to provide an option to turn on acceleration for the RPi4 and any other target that needs it to run properly, using: ./configure --enable-accel=yes
The option defines ACCEL_ON in src/config.h, turns on acceleration in the declarations at the start of src/sdl/video.c and causes "VIDEO_ACCEL=1" to be written into the default .atari800.cfg the first time the emulator is run.
If the option is used for a target that is not OpenGL aware it is ignored.
The default is --enable-accel=no.
I've updated the compilation instructions in DOC/BUILD.RPI.

P.S. While I was doing this I also updated some notes in DOC/README.RPI, then realised they aren't really related to this change.  Hope this is OK - if not, let me know and I will remove them and submit a separate pull request.

mikrosk:
While looking again into this, I have two comments:
- `enable-accel` seems to be way too generic - it says nothing about what it accelerates or for what targets (SDL only in fact)
- how is this option different from `--with-opengl` ? Especially when its description says: `Use OpenGL for graphics - only in SDL target` while yours says: `turns on video acceleration` ? If I were the user, I would be pretty confused. Can't we merge these two into one?

hatchcliff:
I was trying to keep the option name short, but I agree `--enable-accel` isn't meaningful, and its relationship to `--with-opengl` is indeed confusing.

The purpose of the option is to allow the emulator to be built such that video acceleration via opengl is enabled by default on the first run after installation - so I need to find reasonably concise way of expressing that.

I suggest `--enable-opengl-by-default` (or perhaps something even more descriptive like `--enable-opengl-support-by-default-on-installation` would be better - I'm open to suggestions).  The accompanying details would have to be changed too, perhaps like this:

```
A8_OPTION(--enable-opengl-by-default,no,
      [Set VIDEO_ACCEL=1 in the default config file created on first run after installation],
      VIDEO_ACCEL_ON_BY_DEFAULT,[For targets too slow to run without opengl video acceleration,
      supported only with --with-opengl.]
```

I don't think this can be amalgamated with `--with-opengl` because there are two separate decisions to make.  Firstly whether to build with opengl support, and secondly whether to enable that support on installation.  In V4.2.0, support is disabled by default, so there is no second decision to make, but we want to allow an exception for the RPi4.

mikrosk:
Would it be really that bad to make it default for all `--enable-opengl` targets?

hatchcliff:
Three options have emerged from the discussion.  Let me know which one you prefer and I will try to finalise the changes.

1. Instruct RPi4 users to do some manual setup after installation.  They would either need to turn on video acceleration via the configuration menu, or use the runtime option `-video-accel` (which overrides the setting in the configuration file but does not alter it, so would have to be used every time).  This is the simplest option to implement because it requires documentation changes only, but it will catch out some users (perhaps the majority) who do not read documentation prior to routine installation.  They might mistakenly conclude that atari800 does not run properly on the RPi4.

2.  Turn on video acceleration by default for all targets (the original proposal).  This requires a one-line code change to reverse the current default, so is very easy to implement.  The snag is that opengl drivers are known to be broken on some systems, so if we make this change the emulator will crash immediately after installation on some systems that were previously able to run it.  I have no idea how widespread the problem is, but it seems worth avoiding - so this is maybe not the most attractive option.

3.  Turn on video acceleration by default for the RPi4 only (as discussed in recent posts).  This is more complex in terms of coding, but not by much - it involves adding 4 lines of code to configure.ac (actually 1 command) and another 4 lines to src/sdl/video.c.  The benefit is that the setup is done automatically for the user the first time the emulator runs.

It looks like a choice between 1 and 3 to me, but I'm happy to go with the consensus.

joysfera:
3

hatchcliff:
OK, continuing with 3:

I have altered some descriptions and names to clarify the separate options for building and enabling OpenGL support.  `./configure --help` now lists them as follows:

 ```
 --with-opengl           Build support for OpenGL hardware video acceleration
                          - only in SDL target [default=check]

  --enable-opengl-by-default
                          Set VIDEO_ACCEL=1 in the config file created on
                          first run after installation to enable OpenGL
                          hardware acceleration (OpenGL-aware targets)
                          (default set VIDEO_ACCEL=0)

```
The command for configuring the RPi4 version is

`./configure --with-opengl --enable-opengl-by-default`

but because OpenGL support is built by default for appropriate targets this can be abbreviated to

`./configure --enable-opengl-by-default`

mikrosk:
If @joysfera doesn't have any objections, I'll merge it in.

joysfera:
> If @joysfera doesn't have any objections, I'll merge it in.

Go ahead, I'm looking forward to running RPi4 SDL build at full speed :-)

Issue #33, opened 2018-06-08
"Bounty Bob Strikes Back" cart version doesn't recognize the space bar


Shamus397:
This is with the latest version in the git repo and has been an issue for a long time now, going back to v3.1.0 and before. For some reason, the cartridge version of "Bounty Bob Strikes Back" doesn't react when the space bar is pressed on level 2. Also, the lift does not activate when you jump on it either (which it should). Using "1" and "2" on the keyboard to activate the transporters works as it should, so it's not ignoring all keys. This is on an XL NTSC emulated system, using a USB gamepad for input; host system is Linux/SDL.

I also verified that all works normally with the tape version (XL/PAL), so it's just the cart version that has this issue.

mikrosk:
@Shamus397 can you try to reproduce without the gamepad, using keyboard -> joystick emulation only?

mikrosk:
@ChoccyHobNob as you are the local BBSB expert ;), can you confirm this?

ChoccyHobNob:
Sure I’ll give BBSB a good going over today (it’s only 8am here) and let you know about stereo and the space bar

ChoccyHobNob:
OK So, The BBSB works in real hardware even if Stereo POKEYs are installed. I tried multiple versions all of them were fine.

The space bar thing is a different issue to what is being reported. The game has a Utility Hoist (First time you use it is level 2) when you land on the hoist you stick to it, the joystick moves the hoist around, you press space to stop it to get off or press space again to re-stick to it. 

In Atari800 if you land on the hoist it doesn’t get activated at all so you can’t control it. 

This is true for all versions except for the USGold cassette version which works fine and the Homesoft XEX version which is also OK.

Obviously this all works fine on real hardware, also works in Altirra.

There is discussion about it on AtariAge back to 2010 so it’s been a known thing with this emulator for a long time.

There is another issue with BBSB in that it doesn’t work (the joystick is non-responsive) if you have a U1MB installed (The Homesoft XEX is fixed to work, none of the other versions are controllable at all. That made testing all this on real hardware a royal pain as my 1088XEL has a U1MB.

joysfera:
so all we need is to point users to Homesoft XEX version and we're good?

ChoccyHobNob:
ಠ_ಠ

Shamus397:
[...]

PR #123, opened 2020-11-30
fix for BBSB lift not working

mikekov:
This is a fix for "'Bounty Bob Strikes Back' cart version doesn't recognize the space bar" #33 issue.
BBSB cart access fixes for PAGED_ATTRIB and non-PAGED_ATTRIB modes.


joysfera:
Wow, fantastic!

PR #125, opened 2021-01-15
Expand key redefinition

TheRobotFactory:
PR #102 added support for redefining Start/Select/Option/Reset/Quit.

This change adds the remaining F keys to the list of keys redefinable in the config file so that things like F1 as turbo (to match Altirra) can be set up without losing the ability to get to the UI or reconfiguring the S/s/O/R/H keys to the end of the function key row to make them closer to the XE layout but still have access to the later F key functions.


joysfera (approved):
Looks good to me at first glance.

TheRobotFactory:
Is someone going to merge this?

mikrosk:
Sorry for the delay :)

PR #133, opened 2021-05-23
Linux Build Updates

hatchcliff:
This pull request proposes some changes to the debian/rules script and related documentation.  Files modified:

debian/rules
=========
This script routinely generated two warnings:

1. dpkg-shlibdeps: warning: binaries to analyse should already be installed in their package's directory
This has been fixed by asking dpkg-shlibdeps to analyse the copy of the binary file that has been placed the temporary directory for packaging instead of the original.
    • dpkg-shlibdeps src/$(P)	 where $(P)='atari800'
has been changed to:
    • dpkg-shlibdeps debian/tmp/usr/bin/$(P)
      
2. dpkg-gencontrol: warning: -isp is deprecated; it is without effect
The -isp option specifies the new default behaviour, so is no longer required:
    • dpkg-gencontrol -p$(P) --isp
changed to:
    • dpkg-gencontrol -p$(P)

debian/changelog
============
When I did the last two builds, 4.1.0 and 4.2.0, I entered my name as the contact in the changelog, but I have since realised that it is not the package builder's name that belongs here, but the package maintainer, a listed Debian member.  To Antonin Kral: I apologise for this error, I have reinstated your name and relegated mine to a comment alongside the list of packages I built for release on GitHub.

DOC/BUILD.linux_package
=====================
Added an instruction to create an up to date copy of the default configuration file .atari800.cfg for  placement in /debian/DOCS/atari800/atari800.cfg on installation.  I assume this is done to comply with the recommended practice of documenting configuration file changes.  However, we have forgotten to do this on every release since V2.2.0 - which invites the question: does anyone ever use this file?  Let me know what you think.  If the consensus is that the file is not useful, I will remove it from the package and delete the new instruction.

I have also removed some unwanted options from the dpkg-buildpackage command:
    • dpkg-buildpackage -d -B -uc -us
has become
    •  dpkg-buildpackage -B -uc

The -d option turns off checks for dependencies and conflicts.  The procedure contains an earlier test to confirm compilation without error, so this should do no harm, but even so there is no reason to suppress the checks.

The -us option specifies no source package signature, which is irrelevant because we are not building a source package, just a binary one.

DOC/BUILD.RPI
=============
Extended to cover the Pi 400.

Improved description of the --enable-opengl-by-default option.  I realised after a recent discussion on AtariAge that I had not explained its purpose properly.

A few years ago I wrote the section at the end of this document about cross-compiling.  Given that native compilation is quick on the later Pis, I think it is highly unlikely that anyone will resort to cross-compilation in the future.  Hence this section is probably irrelevant - also unmaintainable because of the rapid pace of development of Pi OS.  I have assumed it is worth keeping for the time being as an example of how to use the use of various options and tools associated with cross-compilation, but there may be a case for deleting it - let me know if you have a view on this.

DOC/README.RPI
===============
Extended to cover the Pi 400.

Pi Zero limitations listed.


joysfera:
Thanks. I found the improved README.RPI so useful that I merged the whole PR right away.

PR #136, opened 2021-08-12
AVI video recording

robmcmullen:
This patch adds video/audio recording to AVI format files that are playable with Windows Media Player, all ffmpeg-based players (mpv, vlc, etc.), and likely on MacOS but I no longer have a Mac and can't test there.

Details of patch:

* No new library or external code dependencies
* Video codec used is the simplest video codec I could find that is widely supported (even as far back as Windows XP): the fourcc codec 'mrle', a run-length encoder similar to PCX encoding
* Audio is stored as raw PCM
* Renamed sndsave.c to multimedia.c to reflect that more than just sound can be saved
* Combined video recording with audio recording into new functions like Multimedia_OpenFile rather than SndSave_OpenFile
* Only one of audio or video recording can be active at once (e.g. only Alt-W or the new Alt-V for video recording can be used at one time -- starting one stops the other)
* 4GB file limit with AVI files; this can be worked around but it's more complicated than I wanted to get into for this first patch
* Added "Alt-V" shortcut to record videos

Future ideas:

* Evaluate AVI files for >4GB support. The 'mrle' video codec, when used with complex graphics yield only about 35 minutes of recording time. Best case with simple graphics is about 2 hours of recording. This may be enough for many purposes.
* If >4GB support is too difficult, multiple files could be created that could then be concatenated later using e.g. ffmpeg
* Explore other video codecs to increase recording time. One possibility is to add option to encode video as PNG; this would reduce video sizes by almost an order of magnitude at the expense of being a non-standard codec and only playable using more modern players like ffmpeg, vlc, etc. Specifically, it excludes Windows Media Player.
* Move PCX and PNG image writing from screen.c into multimedia.c to keep external file format support in one place.


joysfera:
I like it, thank you! The 4GB limit will need to be solved, though. I'm sure users will start recording their "speed runs" through games and 2 hours is often not enough given how great games we have on Atari :-)

joysfera:
Comments after testing it for the first time:
- I am not sure if pressing the Enter key after Alt+V is ideal. Especially when the video recording is stopped it seems unnecessary.
- GNOME Totem does not replay the resulting AVI correctly (just a black screen). MPlayer plays it OK. Wondering if YouTube can import the AVI properly?
- it is very handy to be able to create native video recording of Atari800 session. This will be a great new feature for an upcoming Atari800 release.

robmcmullen:
I modeled the Alt+V after the Alt+W command for saving audio, and it does require a return after Alt+W.

I didn't try GNOME Totem, but will see if I can figure out what's going on there. There is an uncompressed fourcc code 'UYVY' that may have wider support, but it uses 4 bytes per pixel and would further limit video size.

I did an upload test to YouTube, and it did successfully convert it. https://www.youtube.com/watch?v=O_nt-1hkyqk

And I forgot to mention: the patch works on big endian machines, too. As a side effect, it also fixes WAV file output on big endian machines.

joysfera:
OK, I never used the Alt+W. Never mind the Totem issue (BTW it's probably using GStreamer as the multimedia framework and I've seen black screen on other videos, too). As long as YouTube handles it we're fine.

Glad you fixed the WAV for big endian.

joysfera:
Forgot to tell you earlier to go for moving the PCX/PNG out from screen.c, that sounds like a good idea. If you could re-use that existing code then even better. Pick the filename as you see it fit.

PR #137, opened 2021-08-17
Improvements to AVI video recording

robmcmullen:
Here's a bunch of improvements to the AVI recording.
* moved PNG & PCX code into new file_export.c file
* moved AVI file creation into file_export.c; kept multimedia.c as high-level driver functions
* Added PNG image compression codec to AVI files
* Added PNG compression level command line argument (affects both screenshots and AVI files)
* added --enable-videorecording configure option to allow compile time selection or exclusion of AVI writing
* Added man pages documenting the video recording and describing new arguments
* Added table in man page showing estimated recording times
* Fixed bug in WAV recording that would allow recording and create zero length file when -nosound option used
* Fixed word alignment bug in AVI files when audio was an odd number of bytes

I put a big table in the man page showing recording times, but the upshot is that with PNG image compression at the highest setting, I'm getting files of 4 hours and more even with the 4GB limit of AVI files. With games that have simple backgrounds, it can even get 6 or 7 hours. Hopefully that will be acceptable in the short term, and if requests come in for longer recording times I will try to look at extensions to the AVI format that get more time, or perhaps another container format.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1075,6 +1075,9 @@ else
 fi
 
 AM_CONDITIONAL([WITH_MULTIMEDIA], test "$with_sound" != no -o "$WANT_AVI_VIDEO_RECORDING" = "yes")
+
+dnl file_export.c is needed when sound is enabled (for WAV recording) or when AVI recording
+dnl is enabled, or for screenshots when screen.c is compiled into the code.
 AM_CONDITIONAL([WITH_FILE_EXPORT], test "$with_sound" != no -o "$have_bitmapped_screen" = "yes")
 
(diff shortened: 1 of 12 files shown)


joysfera:
Very cool, thanks!

robmcmullen:
Hold off for a moment merging this; I made a mistake in the documentation. Turns out that Totem *will* play the PNG encoded videos; it just doesn't play the RLE encoded videos.

robmcmullen:
Woops, sorry, your comment didn't refresh when I refreshed the page. I'll fix the documentation and send another pull request.

PR #138, opened 2021-08-17
Updated video recording documentation for player support

robmcmullen:
Here's the fixed documentation.

I noticed that there is a -screenshots option to specify the filename patterns for screenshots, but no corresponding option for WAV files or the new AVI file generation. Would you like me to add that?


joysfera:
Just unleash your creativity :-)

PR #139, opened 2021-08-18
Allow specification of filenames for sound and video files

robmcmullen:
Added command line options -soundfilename and -videofilename to specify filename pattern like screenshot filename pattern. Moved the filename pattern code into util.c and modified the functions to make them work with the general case, not just screenshots. Changed the default filename patterns to use the hash notation so max value doesn't get out of sync of default pattern ever changes in the code.


joysfera:
Thanks!

PR #140, opened 2021-08-19
configure.ac: Fix compatibility with dash shell

mgorny:
Fix the compatibility of the configure script with the dash shell
by replacing the use of bash-specific '==' operator with plain '='.
The calls in question do not utilize the pattern-matching behavior
of '=='.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -709,7 +709,7 @@ AC_DEFUN([A8_OPTION],[
     fi
 ])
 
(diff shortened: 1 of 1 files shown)


mgorny:
Note that the `${.../.../...}` substitutions are also bashisms but since they're non-trivial to replace and on a Windows-specific code path. I'm not convinced whether they're worth addressing.

joysfera:
Thanks!

PR #141, opened 2021-08-20
Sound support for libatari800

robmcmullen:
Added sound support for the libatari800 targed and included sample code in the test program to write out a WAV file using libatari800 functions.

Also added documentation for the public functions of libatari800 to the README.libatari800 file.


joysfera:
Thanks!

PR #142, opened 2021-08-20
Patch for screenshots and avi videos to respect viewable area

robmcmullen:
Screeshots and AVI videos were using hardcoded values to save the entire Screen_atari array.

This patch modifies the image saving code (affecting both screenshots and the AVI video code) to respect the visible area as defined by the user. For instance, if SDL port is used, image area can be controlled by the -horiz-area & -vert-area command line options, and now the screenshots and videos will show the visible area as defined by those options.

### DIFF
--- a/src/file_export.c
+++ b/src/file_export.c
@@ -33,13 +33,13 @@
 #include "sound.h"
 #include "pokeysnd.h"
 #endif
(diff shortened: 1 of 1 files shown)


joysfera:
Great!

PR #146, opened 2021-09-09
Video codec improvements & libatari800 fixes

robmcmullen:
Added inter-frames to the run-length encoding codec, greatly improving compression ratios. You can record Jumpman for 45 hours! OK, that's without audio, but still over 10 hours with audio in a 4GB file.

Made it the default codec now, because it's compatible with more players, uses less CPU to compress, and compresses better in most cases over the PNG video codec.

Also:
* added `-keyframe-interval` command line option
* added config file support for file_export arguments
* updated man page
* fixed libatari800 key handling
* moved striendswith to util.c so it can be reused by libatari800


joysfera:
Thanks!

PR #147, opened 2021-09-14
ZMBV video codec

robmcmullen:
I had to add one more video codec because the ZMBV codec has **super amazing** compression for like 250 lines of code, and means we can likely ignore the 4GB limit in AVI files. It provides over a week of recording time for Jumpman (with no audio) or a day with audio. It's well supported by players & YouTube. The ZMBV code is modified from the FFmpeg project, which is LGPLv2.1 licensed, permitting modifications to be licensed under the GPLv2. Credits at the top of the file video_codec_zmbv.c.

Other stuff:

* added on-screen display of recording time and the -showstats/-no-showstats command line arg for it (ended up adding the alphabet to the 5x7 screen font. Let me know if that's too much and I can hide some of that behind #ifdefs) Note the on-screen stats are *not* recorded in the video, they are only displayed on the user's screen while recording is happening.
* added configure switches to include/exclude particular video codecs if desired
* updated man page with table of supported video players
* modularized the file_export source code, splitting out video codecs into their own files
* changed -pnglevel argument to -compression-level rather than adding separate -zliblevel argument. One argument is used for both compression level parameters


joysfera:
Wonderful, thanks!

PR #151, opened 2021-09-23
MP3 audio & other audio codecs, and file export code restructuring

robmcmullen:
This patch adds MP3 audio (if libmp3lame is installed) and other audio codecs (mu-law, ADPCM) to have options to reduce video/audio file size at the expense of possible quality loss.

Other stuff:

* restructured file export modules into `codecs/` directory to keep from cluttering up the main source directory
* created classes of codecs: audio, video & image to make new format support easy
* modularized multimedia container support (wav, avi) & added .mp3 file creation
* updated man page with all the new audio options

**NOTE**: libmp3lame doesn't seem to be installed in a standard place, so the include path could vary on different operating systems. I could:

1. include the LAME encoder source in atari800 (it's not terribly big and it's LGPL)
2. include a small GPL'd MP3 encoder (https://github.com/toots/shine) as a fallback if LAME is not installed
3. leave it like it is

Let me know if any of the first two options is interesting!


robmcmullen:
FYI, this is the end of the code restructuring I wanted to do. I'm happy with it; it's super modular, easily extensible, and out of the main directory so it's not mixed with the emulator code and people who don't want it can ignore it.

I don't plan on trying to add every audio/video codec under the sun or anything, so this code isn't going to grow without bounds. With the mp3 audio and ZMBV video, you can record for days, so there's no point in searching for better audio or video codecs for duration.

As a pie-in-the-sky, I would like to add optional support for the mp4 container, the aac audio codec, and the h264 video codec which would provide direct video upload to things like Twitter without transcoding. Also, the m4a container (which is just the mp4 container with a different file extension) would allow creating things like ringtones directly from the emulator.

Those options (mp4, aac, h264) would be optional dependencies requiring configure flags to enable because they would need to pull in huge external libraries.

I need to learn a lot about those topics, though, so it's low priority. I'm going to get back to working on my scriptable ideas and Lua support now. :grin:

joysfera:
Thanks a lot.

PR #152, opened 2021-10-01
Added file export error checking and UI improvements

robmcmullen:
Some minor stuff: error checking and error messages related to mp3 support, better mp3 configuration, small doc edits, etc.

* added --with-mp3 config option to enable/disable mp3 support
* added on-screen error messages if something goes wrong when the user presses Alt+W or Alt+V
* added check for mp3 audio only in mp3 container
* fixed lame.h include file location check in configure.ac
* made .mp3 files be default when saving audio and -acodec mp3 selected
* fixed stats printing when no video frames produced
* added notes in README and INSTALL for mp3 audio support
* updated man page

### DIFF
--- a/DOC/INSTALL
+++ b/DOC/INSTALL
@@ -56,8 +56,13 @@ many others with a little work):-
 11. Static library version: libatari800
 
 Installed zlib (compression library) enables building the emulator with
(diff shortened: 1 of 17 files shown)


joysfera:
Cool, thanks!

PR #157, opened 2022-01-28
Fixed indentation in man page when text follows a list

robmcmullen:
Minor man page fixes to keep a paragraph indented after displaying a list.


joysfera:
Thanks!

PR #162, opened 2022-06-08
remove rpi builder

hatchcliff:
The released v5.0.0 Raspberry Pi package (atari800_v5.0.0_rpi.deb) contains two binaries, one of which is selected at run time.  It was intended that this single package would serve all models of Raspberry Pi, replacing the pair of packages released previously.  This arrangement has not been well received by the users, and there is no support for its retention.

This pull request effectively reverts #149.  It removes the raspbian folder used to create the new package, and its associated documentation file DOC/BUILD.linux_rpi_package.  One icon (data/atari800rpi.png) and a minor change to DOC/BUILD.linux_package have been retained.

The debian folder that was used to build Pi packages prior to v5.0.0 is unaffected by this pull request, and will return to use for future releases.


hatchcliff:
It looks as though the whole issue of multiple binaries/packages has become irrelevant in Pi OS Bullseye, strengthening the case for #162.  There are two reasons for this:

-  Bullseye contains a new standard KMS video driver which supports the stock Linux version of Atari800 on a good range of Raspberry Pi models.  So a single package may suffice for future releases.
- Some of the Pi-specific graphics libraries previously used in the special 'Pi Zero, 2 and 3' package have been dropped from Bullseye.  The --target=rpi configuration option that was used to build this is now a legacy feature that only works on the earlier Stretch and Buster versions of Pi OS.

There is an ongoing discussion about this on AtariAge:
https://forums.atariage.com/topic/336269-installing-atari800-v500-on-the-raspberry-pi/?do=findComment&comment=5117109
I have posted an experimental package for others to try out, and am currently awaiting feedback to confirm its suitability for future release.

Issue #166, opened 2022-09-21
Feature request: Allow continuing of the emulation after playback file finishes


marcin-jozwikowski:
It would be a useful feature to allow for the emulator to continue working as usual after the playback file is done.

It would work as a macro by just replaying the beginning of something (setting up, loading, replaying first levels of a game) and then allowing the user to take control back into their hands.

mikrosk:
Fixed by #167.

PR #167, opened 2022-09-21
Issue #166 - Continue emulation after playback

marcin-jozwikowski:
* Added the -playbacknoexit flag to commandline parameters of src/input.c
* Added flags' description in DOC/USAGE and src/atari800.man
* Implemented flag functionality by conditionally skipping emulation exit
* Put -help records values for playback related featuren under ifdef condition to remove them when the feature is disabled in build params


mikrosk:
Looks good, merged. Thanks!

PR #169, opened 2022-10-17
Fixed #168: added ifdefs to correct build errors with --disable-videorecording

### LINKED ISSUE #168: Unable to build with --disable-videorecording
I guess somewhere is missing an `#ifdef VIDEO_RECORDING` but I'll let @robmcmullen to decide.

robmcmullen:
Added missing conditional defines, also verified compilation with combinations of --with-sound=no and --enable-videorecording.

### DIFF
--- a/src/codecs/container.c
+++ b/src/codecs/container.c
@@ -241,6 +241,7 @@ int CONTAINER_Open(const char *filename)
 	return (fp != NULL);
 }
 
+#ifdef SOUND
 int CONTAINER_AddAudioSamples(const UBYTE *buf, int num_samples)
 {
 	int result;
@@ -257,7 +258,11 @@ int CONTAINER_AddAudioSamples(const UBYTE *buf, int num_samples)
 			return 1;
 		}
 	}
-	else if (!video_codec) {
+	else
+#ifdef VIDEO_RECORDING
+		if (!video_codec)
+#endif
+	{
 		/* Before file close time, there is one call to this function every
 		   video frame. If the video codec is not being used, we need to count
 		   frames here because frame count is used to determine the duration of
@@ -304,7 +309,9 @@ int CONTAINER_AddAudioSamples(const UBYTE *buf, int num_samples)
 	}
 	return result;
 }
+#endif
 
+#ifdef VIDEO_RECORDING
 int CONTAINER_AddVideoFrame(void)
 {
 	int size;
@@ -356,6 +363,7 @@ int CONTAINER_AddVideoFrame(void)
 
 	return result;
 }
+#endif
 
 /* Closes the current container, flushing any buffered audio data and updating
    the container metadata with the final sizes of all video and audio frames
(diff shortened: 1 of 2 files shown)


mikrosk:
My ultimate goal was ` --disable-videorecording --disable-pokeyrec` but IIRC the latter worked ok. I'll give it a try today, thanks!

mikrosk:
Merged, thanks again.

Btw, would it be too much to ask to make the whole file export feature configurable? In Atari Falcon port (and perhaps some others) there is no need for such powerful features and the binaries gets fatter for no real benefit even with `--disable-videorecording --disable-pokeyrec`

robmcmullen:
I was thinking about that, perhaps adding a `--disable-file-export` config option that would exclude all that code. What about saving to PCX/PNG? I assume PNG is not likely on the 68030 anyway, but do you use PCX image export? Or should image export be separate from audio/video exporting?

mikrosk:
Maybe it is enough just to have `--disable-audiorecording` which would hide everything audio-related (i.e. mainly the built-in pcm, adpcm and mulaw code + perhaps some handlers) and hide PNG/PCX under `--disable-videorecording` (if I understand it correctly, that PCX export is meant as storing PCX images each frame?)

Those export functions are useful but not really on low-end hardware, there you are happy that it runs, you're not going to export anything from it. :D

Also, what is the relationship between `--enable-pokeyrec` and our hypothetical `--enable-audiorecording` ? Is the former something totally different?

robmcmullen:
PCX/PNG is for screenshots, so perhaps there should be a separate option to disable that if Falcon or low-end platforms could benefit from excluding that code.

`--enable-pokeyrec` seems entirely separate, it looks like a debugging interface to write pokey audio values.

I'm happy to add this stuff, because as of now, the audio recording stuff is enabled if SOUND is defined and it should be a separate config option.

robmcmullen:
Ok, I'll implement `--disable-file-export` that will act as a big switch to turn off the multimedia/screenshot options. I'll also add `--disable-audio-recording` and add a config var so that the audio recording code doesn't use `SOUND` any more, as well as `--disable-screenshots` for completeness.

joysfera:
I'm not exactly excited by such polluting the source code by more and more ifdefs. Is it really necessary, @mikrosk ? Who's going to use them to disable anything?
BTW, if I were an active Falcon user I wouldn't be happy if I got severely crippled Atari800 binary. That would rather force me to switch the host platform than to keep it.
I mean @robmcmullen go ahead if you're willing to add this big switch but frankly I can't imagine a platform that would benefit from disabling these nice features we've got in Atari800.

robmcmullen:
I can understand not wanting to pollute the code. I think I did make a mistake in using `SOUND` as the #define that triggers audio recording capability, because maybe there are other cases where a platform might want sound without recording capability? I'm happy to correct at least that. But I'll let you two decide how far I should go.

joysfera:
In this particular case go as far as you're willing to but in future I'd consider adding more ifdefs to the main parts of the code. Let's keep it readable at least.

mikrosk:
> Who's going to use them to disable anything?

Me, for starters. Binary size in 5.0 increased in comparison to 4.2 by sometimes as many as 70 KB and for no real benefit for the Falcon machine

> BTW, if I were an active Falcon user I wouldn't be happy if I got severely crippled Atari800 binary. That would rather force me to switch the host platform than to keep it.

Exactly! You want to record video and audio on the host platform, not on Falcon. For the same reason we don't deliver the extended monitor functionality on Falcon, it's useless.

So from my perspective, such switch if much appreciated. If you are afraid of too many options, I'm totally fine with just one global switch (`--disable-exports` or something like that) which would disable everything (audio, video, pictures).

PR #171, opened 2022-10-20
Added --disable-screenshots and --disable-audiorecording

robmcmullen:
Audio recording is now uncoupled from the `SOUND` constant so you can still have sound but not include the audio recording capability. Also added the `SCREENSHOTS` constant tied to the `--[dis|en]able-screenshots` configure argument. Both these constants also clean up the `#ifdef` blocks, because there were tests like `#if !defined(BASIC) && !defined(CURSES_BASIC)` that can now be replaced by tests like `#ifdef SCREENSHOTS`.

I decided not to add a `--disable-file-exports` because there are other things like save states that could be considered as exporting a file but that's not what was intended here. Hopefully adding `--disable-screenshots --disable-audiorecording --disable-videorecording` isn't too bad for the few platforms that are so space constrained that they would exclude all these super mega awesome features :smile:, because by default all those features are enabled.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -362,7 +362,7 @@ else
 	AC_CHECK_LIB(z,gzopen)
 
     dnl Need SUPPORTS_LIBPNG shell variable for PNG video codec test below
-	AC_CHECK_LIB(png,png_get_libpng_ver,[LIBS="-lpng $LIBS"; SUPPORTS_LIBPNG=yes],[SUPPORTS_LIBPNG=no])
+	AC_CHECK_LIB(png,png_get_libpng_ver,[SUPPORTS_LIBPNG=yes],[SUPPORTS_LIBPNG=no])
     if [[ "$SUPPORTS_LIBPNG" = "yes" ]]; then
         AC_DEFINE(HAVE_LIBPNG,1,[Supports PNG image compression library.])
     fi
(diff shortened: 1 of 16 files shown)


mikrosk:
There's a couple of suspicious constructs in `cfg.c`:

```
#if defined(AUDIO_RECORDING) || defined(VIDEO_RECORDING)
#include "file_export.h"
```
OK, this makes sense, we want call the export code only for audio and/or video.

```
#if defined(HAVE_LIBPNG) || defined(HAVE_LIBZ) || defined(AUDIO_RECORDING) || defined(VIDEO_RECORDING)
			else if (File_Export_ReadConfig(string, ptr)) {
			}
```
Why is this different than the include file? Also, does libpng || libz represents the screenshot functionality? If so, it should be noted somewhere.

```
#if (defined(SOUND) && !defined(DREAMCAST)) || defined(VIDEO_RECORDING)
	File_Export_WriteConfig(fp);
#endif
```
Something different again from the above and yet referencing the same include file.

I haven't looked further, it's possible there is more inconsistencies like this.

robmcmullen:
Thanks for reviewing this, I'll fix that and look around for more things I may have missed.

robmcmullen:
Pls review these fixes:

1. fixed the `DREAMCAST` problem; just missed that one.
2. found some code in `file_export.c` that should not be included if only `libz` is defined.
3. `file_export.h` must be included in `cfg.c` when any of the following happen: audio or video recording, `libz` being used, or `libpng` being used. The latter two are because there's a `-compression-level` argument that allows a space vs time tradeoff for those libraries. Screenshots doesn't automatically require `file_export.h` in `cfg.c` because if `libpng` is not used and audio/video are not used, there are no configuration options that apply. Only when `libpng` is used with screenshots does `file_export.h` become necessary.

mikrosk:
Shouldn't `File_Export_WriteConfig` in `cfg.c` be also surrounded by `defined(HAVE_LIBPNG) || defined(HAVE_LIBZ)` ? It seems to depend on those options exactly as you write - for that compression level setting.

Apart from that it looks good, Falcon builds with your changes are now about 30 - 50 KB smaller than the 4.2.0 release.

mikrosk:
Merged, thanks!

PR #179, opened 2023-01-08
Fixed keystrokes for inserting/deleting of line/character

mirao:


mirao (commented):
I tested the key shortcuts on Linux (Ubuntu 22.10) with Atari 800 emulator 5.0.0, machine Atari 800XL

PR #186, opened 2023-06-26
Update AltirraOS to 3.34, build 5200 ROM from source

asiekierka:


mikrosk:
Merged, thanks!

dmsc:
Hi!

Note that current Altirra OS it at version 3.38, see my repo at: https://github.com/dmsc/atkernel

Have Fun!

asiekierka:
Sure - but what about Altirra BASIC? I tend to use the latest stable Altirra release as my update source, as that lets me update everything in one go to matching versions; but I'm willing to use a newer source, if one exists.

dmsc:
Hi!

I don´t track Altirra BASIC in my repo because it does not change often - last change is 1.58 from December 2021 (at Altirra 4.10-test3), IMHO it would be better to prepare a different repo for it.

Current Altirra OS version (3.38) is included in Altirra 4.20-test15, from a week ago.

And the 5200 kernel is unchanged from the re-licencing on 2016 (at Altirra 2.80).

Have fun!

PR #192, opened 2023-09-26
Rpi debian tidy

hatchcliff:
Previously we have released two packages for the Raspberry Pi, a stock Debian one for the Pi 4; and a special one for the Pi Zero, 2 and 3, built using the --target=rpi configuration option.

Since the release of Pi OS Bullseye this is no longer necessary, because Bullseye contains a new standard KMS video driver capable of supporting the Debian version on a wide range of Pi models.

Some of the Pi-specific graphics libraries previously used in the special version have been dropped from Bullseye, so the --target=rpi configuration option is now a legacy feature that only works on the earlier Stretch and Buster versions of Pi OS.

This pull request changes configuration and packaging scripts, documentation files, icons and program comments only - no executable code.

I've added a new icon (data/atari1.png), because the existing one (data/atari2.png) only works well on desktops with dark themes.  The new one works well on both light and dark themes.

I've also deleted two disused files: the former Raspberry Pi icon data/atari800rpi.png and the reference-only copy of the configuration file debian/atari800.cfg which has been out of date in every release since v2.2.0 and appears to be an unnecessary maintenance burden.


joysfera:
Thank you!

mikrosk:
Looks good to me, just one question:

> At the time of writing (September 2023) it does not run properly on Pi OS Lite.
> It is not known whether this is a temporary situaton that will be corrected by
> further development, or if new software such as the KMS video driver that atari800
> depends upon to run at full speed have been deliberately omitted from the Lite version.

From this snippet it is unclear to me whether the reason why it doesn't run on the Lite version  is omitting the KMS driver _for sure_ or whether it is only your guess.

hatchcliff:
It's just a guess.

When I first tried this about a year ago I found that the KMS driver was not installed by default in Lite, but could be easily added using raspi-config.  I think it was labelled an experimental version.  It did not work properly with atari800 (I can't remember exactly what was wrong).

When I tried again today with a fresh system, I found that raspi-config no longer offers the KMS install option, so I'm guessing it has been withdrawn.  I searched for clarification but couldn't find anything.

PR #197, opened 2023-11-18
Allow using Keyboard and SDL controllers as separate joysticks

gagern:
Having [read about M.U.L.E.](https://www.filfre.net/2013/02/dan-bunten-and-m-u-l-e/), I wanted to give that a try with a group of 4 players. 3 controllers and 1 keyboard, or so I hoped. But it didn't work out. And it took considerable working with the code and some recorded inputs to figure out that both the keyboard and the first SDL device were controlling the emulated joystick 0, while nothing was controlling the emulated joystick 3. Pity.

So I wrote some code to change that. This pull request introduces a new command line flag `-joy-distinct` which places the physical SDL devices *after* the keyboard-emulated devices, based on which keyboard-emulated devices are enabled at the time the config file and command line flags have been processed. This doesn't achieve full control over freely mapping which combinations of physical devices should control which emulated joysticks at which point in time, but it's a start and good enough for my use case. Extended logging output will make the assignments clearer and hopefully help future users who might get surprised by the default behavior the way I was surprised.

Reading the existing code I found that a lot of places were making very non-trivial assumptions about which kinds of host devices could be associated with which emulated controller. Different arrays were linked by index in various places. The possibility of having LPT joysticks and an on-screen keyboard (OSK) thrown in didn't make things any simpler. So the first commit of this request is a fairly wide-ranging change to simplify the abstraction of an emulated device without introducing any actual changes in observed behavior. A later commit then performs the changes based on the setting of the flag. One commit is a drive-by improvement by means of bit field arithmetic which is not really related to this PR except that I happened to look at that code and see a potential for simplification. Readers may choose to look at the commits in isolation to separate these aspects. The commit descriptions have more details which I won't repeat here.

### DIFF
--- a/DOC/CREDITS
+++ b/DOC/CREDITS
@@ -443,3 +443,6 @@ Marcin Zukowski <eru@ibb.waw.pl>
     - fix in ANTIC, GTIA
     - slight monitor improvement
     - fix in Atrax cartridge bank switching
+
+Martin von Gagern <martin@von-gagern.net>
+    - allow SDL joysticks to come after keyboard-emulated joysticks
--- a/DOC/USAGE
+++ b/DOC/USAGE
@@ -202,7 +202,6 @@ Command line options
 -multijoy             Emulate MultiJoy4 interface
 -directmouse          Use mouse's absolute position
 -cx85 <num>           Emulate CX85 numeric keypad on port <num>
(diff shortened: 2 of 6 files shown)


mikrosk:
Looks good, merged. Thanks for the attention to detail!

Issue #156, opened 2022-01-16
AltL-J key to swap (real) joystick in sdl port doesn't do anything


TauPan:
With recent atari800 from github 52af3c69d7ab91fd6707c691fe767a99e60b6b36 Nov. 6, compiled on a DevTerm (raspberry pi with arm64) via `fakeroot debian/rules binary` start ballblazer in atari800, with internal gamepad as /dev/input/js0 and speedlink competition pro extra usb as /dev/input/js1, moving the axis on the first joystick will always move the top player and the other one the lower player, no matter how desperate I press AltL-J.

(Instead of ballblazer ` 10 print stick(0), stick(1): goto 10`  in basic. I guess most people will have ballblazer...)

(The use-case is of course to play boulder dash with the competition pro instead of the d-pad.)

Or is that only valid for the keybboard joystick?

gagern:
During my work on https://github.com/atari800/atari800/pull/197 I noticed that indeed the swapping of joysticks only applies to the keyboard-emulated ones. Looking for example at the code around

https://github.com/atari800/atari800/blob/c0f94a7bf15543d8feb87ba046c456087c9d8227/src/sdl/input.c#L1590

you see that the keyboard input comes before, but the SDL input comes after the swap.

I have some ideas how physical joysticks could get swapped as well, and I personally believe that would make sense and shouldn't do any harm. But as https://github.com/atari800/atari800/pull/197 entails some significant code changes I'd like to see that merged first. Also I can't promise that I'll find the time to write code for this.

mikrosk:
@gagern feel free to go ahead with this one. :-)

joysfera:
Alt-J was supposed to switch keyboard joysticks only, because physical joysticks can be simply handed over, no need to switch them in software.

TauPan:
Switching physical joysticks is hard on my devterm which has a built in d-pad but I'd prefer to use the competition pro as primary joystick. Doing so requires shenanigans with the udev devices which I failed to automate so far.

Sorry for not stating the use-case more specifically.

PR #200, opened 2023-11-24
Also swap SDL joysticks on AltL+J

gagern:
This swaps all physical devices connected to the first two emulated joysticks: the keyboard keys, the SDL devices, and the LPT devices.

After the swap the real joystick / hat setting in the UI will appear swapped as well, but the keyboard setting won't.  Changing that would require one more indirection. I might do something like that as part of a revamped UI integration in a future PR.

The PR includes two drive-by improvements which are not strictly related to the feature in question. See their commit messages.

Fixes https://github.com/atari800/atari800/issues/156.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@
 /src/android/proguard-project.txt
 src/guess_settings
 src/libatari800_test
+/tools/cart
 /config.log
 /config.status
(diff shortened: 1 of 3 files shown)


mikrosk:
Looking good, thanks!

PR #201, opened 2023-11-29
added xex reading in monitor (useful for patches)

jhusak:
This helped a lot with game patch making.


joysfera:
Nice one, thanks!

Issue #199, opened 2023-11-21
BBSB warp to screen 4 does not work


Shamus397:
When you clear level 5 in Bounty Bob Strikes Back! (cartridge version) you get a message telling you how to warp from screen 1 to screen 4:

"IN LEVEL 1, IMMEDIATELY AFTER PICKING UP THE FLOWER POT, HOLD DOWN THE NUMBER 1 ON THE KEYBOARD AND PRESS START TO WARP TO LEVEL 4."

However, this doesn't work; if you follow the instructions (pick up the flower pot, hold down "1", then press START), nothing happens.  Just for the sake of completeness, I checked the tape version of BBSB and it suffers from the same problem.

All on a stock NTSC 800XL--nothing fancy added.

mikrosk:
Does this work in e.g. Altirra?

phaeron1024:
It does. I ran an execution trace in Altirra and this is the code that detects the 1+Start combination:

```
    7010:  9: 61 | A=0E X=01 Y=01 S=FF P=77 ( V IZC) |     0E0B: AD 5F D2          LDA $D25F    ;SKSTAT
    7010:  9: 65 | A=FB X=01 Y=01 S=FF P=F5 (NV I C) |     0E0E: 29 04             AND #$04
    7010:  9: 67 | A=00 X=01 Y=01 S=FF P=77 ( V IZC) |     0E10: D0 ED             BNE $0DFF
    7010:  9: 69 | A=00 X=01 Y=01 S=FF P=77 ( V IZC) |     0E12: AD 59 D2          LDA $D259    ;KBCODE
    7010:  9: 73 | A=1F X=01 Y=01 S=FF P=75 ( V I C) |     0E15: 29 3F             AND #$3F
    7010:  9: 75 | A=1F X=01 Y=01 S=FF P=75 ( V I C) |     0E17: A8                TAY
    7010:  9: 77 | A=1F X=01 Y=1F S=FF P=75 ( V I C) |     0E18: B9 E9 AA          LDA $AAE9,Y  ;$AB08
    7010:  9: 82 | A=11 X=01 Y=1F S=FF P=75 ( V I C) |     0E1B: 38                SEC
    7010:  9: 84 | A=11 X=01 Y=1F S=FF P=75 ( V I C) |     0E1C: E9 11             SBC #$11
    7010:  9: 86 | A=00 X=01 Y=1F S=FF P=37 (   IZC) |     0E1E: 85 AD             STA $AD
[...]

PR #203, opened 2023-11-30
Fix for issue #199 (BBSB warp doesn't work)

Shamus397:
Fix for issue #199 

N.B.: This only fixes the SDL and X11 backends; other backends are left as an exercise for the reader.  ; )


joysfera:
Thanks!

PR #204, opened 2023-11-30
Added H: device rename; save it in setup file

jhusak:
I have added the commandline arg: -Hdevicename <X>, for example -Hdevicename D - which let me run a game that is a DOS dependent without DOS.
Only first non-space character will be a device name.
However, there is completely no checking what user entered.  But the same is in other options, so I let it stay like that. In case the device exists in HATABS, the duplicate will be added.

In the future, an option in menu may be added.

### DIFF
--- a/src/devices.c
+++ b/src/devices.c
@@ -626,7 +626,7 @@ int Devices_Initialise(int *argc, char *argv[])
 				Log_print("\t-H3 <path>       Set path for H3: device");
 				Log_print("\t-H4 <path>       Set path for H4: device");
 				Log_print("\t-Hpath <path>    Set path for Atari executables on the H: device");
-				Log_print("\t-Hdevicename <X> Rename H: device for something different");
+				Log_print("\t-Hdevicename <X> Use this letter to access the Host device, instead of H:");
 				Log_print("\t-hreadonly       Enable read-only mode for H: device");
 				Log_print("\t-hreadwrite      Disable read-only mode for H: device");
 				Log_print("\t-devbug          Debugging messages for H: and P: devices");
--- a/src/devices.c
+++ b/src/devices.c
@@ -496,6 +496,9 @@ int Devices_h_read_only = TRUE;
    is used. */
 char Devices_h_exe_path[FILENAME_MAX] = DEFAULT_H_PATH;
 
+/* H device rename; one can add 'D' in command line */
+char Devices_h_device_name = 'H';
+
 /* Devices_h_current_dir must be empty or terminated with Util_DIR_SEP_CHAR;
    only Util_DIR_SEP_CHAR can be used as a directory separator here */
 char Devices_h_current_dir[4][FILENAME_MAX];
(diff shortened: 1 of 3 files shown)


src/devices.c
```diff
@@ -617,6 +626,7 @@ int Devices_Initialise(int *argc, char *argv[])
 				Log_print("\t-H3 <path>       Set path for H3: device");
 				Log_print("\t-H4 <path>       Set path for H4: device");
 				Log_print("\t-Hpath <path>    Set path for Atari executables on the H: device");
+				Log_print("\t-Hdevicename <X> Rename H: device for something different");
```
dmsc, line 629: IMHO, a better description could be like "Use this letter to access the Host device, instead of H:".

mikrosk:
> In the future, an option in menu may be added.

@joysfera what is (was :)) the policy for adding command line parameters? Does every parameter have to an UI counterpart or it is OK to introduce standalone command line switches?

src/devices.c
jhusak: How to do this best? As second commit or create new pull request with one commit? I am a kind of new in PullRequesting - what is the procedure with PR which have to be changed/fixed?

jhusak:
However, I have commited this change. If I have to make this one commit, write me please. The second option is to squash two commits, but I do not have Github Desktop.

jhusak:
ok, thanks, it makes things much simplier :) The PR is ready :)

jhusak:
Well next PR is waiting (cleanup rather)...

joysfera:
@mikrosk it's indeed better when a new feature comes with UI setting as well. But in this case we might want to test this from command line before bothering with UI refactoring, to find out how useful it is.

jhusak:
BTW, the menu is hard to develop. That stopped me to do the menu entry for this option; I was afraid I'd miss something.
And I needed it asap, so that was a good start to check if this is valuable and usable for people (I think: yes, it is)
I'll think about changing the menu structure to more struct-driven than code-driven. Once I have done this for microcontrollers and LCD.

jhusak:
About settings H: rename I propose such solution:
The line:
"H: device (hard disk)"
change to:
"Hard disk device (<renamed letter>:)" , eventually "<renamed letter>: device (hard disk)"

Lines H1:...H4: rename so dynamically

and everywhere H: is mentioned , renamed letter should appear.

In advanced options add line:
Use hard disk device as '<letter'. 

When pressed, add requester with a letter to enter from keyboard (or a mesh with all A-Z letters?)

What do you think?

mikrosk:
@jhusak I think entering the letter is sufficient, I guess everyone knows the alphabet. ;-)

The H: options could use some cleanup. I'd put everything into "Hard disk options". Then you could have:

- Hard disk device:  H (changed after clicking)
- Device state: disabled / ro / rw
- Atari executables path
- x1 ... x4 clickable paths (no idea why these are duplicated in the advanced options and not even clickable)
- is that "currently open files" good for something?

jhusak:
It is rather not to enter !@#$%& as device "letters".

dmsc:
Hi!

I created a pull request with a proposed new menu arrangement.

> @jhusak I think entering the letter is sufficient, I guess everyone knows the alphabet. ;-)

Instead, I implemented that each click advances the current letter, avoiding "C", "E", "K" and "S". It
is simpler to implement and fast for the user.

> 
> The H: options could use some cleanup. I'd put everything into "Hard disk options". Then you could have:

I also renamed "hard disk" to "host device". IMHO, this makes sense, as nobody actually uses hard-disks
as "H:" in real Atari, and the functionality actually exports "host" files to the emulator.

> 
> * Hard disk device:  H (changed after clicking)
> * Device state: disabled / ro / rw

I put the enable/disable first, because the other options "depend" on it.

> * Atari executables path
> * x1 ... x4 clickable paths (no idea why these are duplicated in the advanced options and not even clickable)

Those are the current working directory in each of the emulated devices.

> * is that "currently open files" good for something?

I renamed the menu to "status", as those are actually the status of the H devices.

Have Fun!

Issue #177, opened 2022-12-27
CTRL Key not working on macOS (MacPorts and Brew)


krischik:
The CTRL key is not working macOS. This is the case for both the MacPorts as well as the Brew distribution.

# Reproduce

Start emulator into Atari-Basic. Type:

```
10 PRINT "[CTRL]Q[CTRL]W[CTRL]E[CTRL]R[CTRL]T[CTRL]Z"
RUN
```

(I have a german keyboard, hence the `[CTRL]Z` instead of `[CTRL]Y`)

# Expected:

<img width="1462" alt="image" src="https://user-images.githubusercontent.com/685853/209652885-11a4745d-6205-4e1d-8bd2-6b7b954f4c14.png">

# Observed:

<img width="1698" alt="image" src="https://user-images.githubusercontent.com/685853/209653258-e61621f2-f08d-4c26-9447-548e4ab4e05d.png">

# Hint:

The **atari800** version from @atarimacosx works correctly. AFAIK @atarimacosx uses SDL Library 2.0 and Metal API and is based **atari800** Version 4.2.

atarixle:
What version of Mac OS X / OS X / macOS are you running?
May be it has to do with SDL1. SDL1 is not working well anymore on equal / anything higher than High Sierra. This is where sound issues in Atari800MacX started (I heard this from Mac owners who had these OSses in use, while I mad the jump from 10.7 to 10.11, and then to 10.14, 11, 12 on patched Macs).
When I built Atari800 5.0.0 myself on 10.4, 10.5, 10.6, 10.7, 10.8, and 10.11, I didn't observe this issue, these systems fully support SDL1.

krischik:
The bug is from December. Before I answer let me update both Brew and MacPorts and check if the bug still exists.

dmsc:
Please, test the patch attached - I also experienced this bug when Debian switched from SDL1 to SDL12-COMPAT.

[001-fixes-keyboard-with-sdl12-compat.txt](https://github.com/atari800/atari800/files/12663393/001-fixes-keyboard-with-sdl12-compat.txt)

This is also in a branch in my repo: https://github.com/dmsc/atari800/commit/4e02766fb9ac05d98449f8cb3cfb5731c66524ca

jhusak:
The patch worked on Ventura/M1. Thank you.

joysfera:
@dmsc would you create a pull request with the keyboard fixes? Has it been tested on more platforms yet?

[...]

PR #205, opened 2023-12-01
Fixes CTRL and CAPS keys when using SDL12-compat library.

dmsc:
This also simplifies the code, as the new SDL key handling is saner. Fixes atari800/atari800#177

### DIFF
--- a/src/sdl/input.c
+++ b/src/sdl/input.c
@@ -437,16 +437,6 @@ int PLATFORM_Keyboard(void)
 	static int resize_w, resize_h;
 #endif /* HAVE_WINDOWS_H */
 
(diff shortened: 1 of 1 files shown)


krischik:
Great. Now all we need is the 5.1.0 release.

mikrosk:
@joysfera likes to do the release on [Atariada](https://atariada.cz/) so give it a few months. :-) In the meantime I might take a look at some of the other bugs as well.

joysfera:
Actually I like to make a release whenever I find few spare hours. Christmas vacation is also a very good candidate, so @mikrosk hurry up with fixing some of the other bugs as a release might be real near! :-)

PR #208, opened 2023-12-08
Add menu option to set 'H' device letter.

dmsc:
Also, rename 'hard disk' to 'host device', that better describes what this device does, and reorder the options to logically group together the 'H' ones.

This is a WIP for the discussion at pull #204

### DIFF
--- a/src/ui.c
+++ b/src/ui.c
@@ -2212,7 +2212,6 @@ static void AtariSettings(void)
 		UI_MENU_CHECK(3, "SIO patch (fast disk access):"),
 		UI_MENU_CHECK(17, "Turbo (F12):"),
 		UI_MENU_CHECK(19, "Slow booting of DOS binary files:"),
-		UI_MENU_ACTION(4, "H: device (hard disk):"),
 		UI_MENU_CHECK(5, "P: device (printer):"),
 #ifdef R_IO_DEVICE
 #ifdef DREAMCAST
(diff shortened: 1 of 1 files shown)


mikrosk:
Thanks! At a first glance it looks good, I'll take a look later today.

mikrosk:
@joysfera ah, so this what you meant. You kept this one open intentionally and not merged it into the 5.1.0 release. As far as I'm concerned, you could have merged it. Maybe you can do a quick 5.1.1 release. :)

mikrosk:
@dmsc since it has _not_ been merged yet (I apologise for my slow response) I have a few suggestions:
- the way how user selects the H: (etc) letter is quite irritating when you miss "your" letter: you have to go through the whole alphabet again. I'd suggest to either list all letters in as a combo box or to allow the user to enter the letter and do some validation (i.e. just A-Z)
- the H: section takes way too much space. I think it's better to put it under one menu item and then list all of the sub-items. This feature isn't so widely used, there's no need for it to take 1/3 of the screen space
- is that "Current directories:" really needed? What is it good for? Why not just show the mapped paths?

Other than that, I like it, very nice cleanup.

joysfera:
I'll give @dmsc few days to see if he can incorporate your suggestions.

dmsc:
Hi!

> @dmsc since it has _not_ been merged yet (I apologise for my slow response) I have a few suggestions:
> 
> * the way how user selects the H: (etc) letter is quite irritating when you miss "your" letter: you have to go through the whole alphabet again. I'd suggest to either list all letters in as a combo box or to allow the user to enter the letter and do some validation (i.e. just A-Z)

Will try to implement something, the UI stuff is not that advanced.... 

> * the H: section takes way too much space. I think it's better to put it under one menu item and then list all of the sub-items. This feature isn't so widely used, there's no need for it to take 1/3 of the screen space

I will put all the options in the current "advanced" sub-menu.

> * is that "Current directories:" really needed? What is it good for? Why not just show the mapped paths?

Don't know if it is useful, I have never used it :-P
It is *not* the mapped paths, it is the folder *inside the path* that is currently selected in the emulated Atari.

> 
> Other than that, I like it, very nice cleanup.

Ok, will update the pull request in the afternoon.

Have Fun!

dmsc:
done, please review!

joysfera:
@dmsc Looks good, thank you! Wondering what the "Current directories" should show, I got them all empty.

BTW, this is not the best place to ask but anyway: does someone remember why we seem to no longer support the H(index+4) virtual devices that converted Atari EOL to host EOL (and possibly other characters) on-the-fly?

LIST "H1:TEST.LST"
LIST "H5:TEST2.LST"

I get ERROR- 160 on the second command.

mikrosk:
Ok, this looks indeed much better. I have noticed only one thing, the input isn't validated. So one can enter drives like `3:` or `$:`. Not a showstopper but should be validated.

joysfera:
I documented in the NEWS file that only letters could be used :-) 
Thus I agree we need to add the validation eventually.

Another "prohibited" device letter could be "R:", BTW... I think it's used for networking?

mikrosk:
There's more, e.g. `P:` for host printer. And then of course is the question what happens if you run a cartridge which install `T:` and you just set your hard drive to be T:.

dmsc:
Hi!

> I documented in the NEWS file that only letters could be used :-) Thus I agree we need to add the validation eventually.
> 
> Another "prohibited" device letter could be "R:", BTW... I think it's used for networking?

Yes, that was a mistake - I intended to validate to working letters.

I will settle for this list of valid device letter: " A B D F G H I J L M N O Q T U V W X Y Z ".

dmsc:
Ok, created a new pull request.

PR #209, opened 2023-12-16
Raspberry PI System Specification

hatchcliff:
I've updated the Raspberry Pi system specification to describe performance with the recently released Bookworm version of Pi OS.

The new specification has been publicised on [AtariAge](https://forums.atariage.com/topic/336269-installing-atari800-v500-on-the-raspberry-pi/?do=findComment&comment=5340538) and no problems have been reported - so everything is running as intended as far as I can tell.

I've also corrected a minor formatting error in configure.ac that I made in a #192.  I used tabs for indentation, but have now replaced them with spaces for consistent display in different editors.


joysfera:
Thanks :+1:

Issue #188, opened 2023-07-16
Monitor / Debugger: Keep Atari screei displayed / show scan line


craigmaloney:
First of all, thank you for an amazing emulator.

I was wondering if there was a way to keep the Atari's screen on while engaging the monitor? Even better would be to show the current scan line / CRT gun position while displaying the image. Something similar either to how Altirra or Stella work. This would be a huge improvement to my development  workflow and understanding of the Atari system in general.

Thank you!

mikrosk:
That's not so hard to achieve, actually. I guess the interesting question would be how to design it. As you have certainly noticed, right now the UI completely covers the screen (and pauses emulation). Maybe there could be a shortcut not to return immediately to emulation but into some sort of "info screen" about current image.

joysfera:
with SDL2 we should be able to open up the monitor/debug in a separate window.

mikrosk:
If I'm not mistaken, @craigmaloney speaks about _graphics_ debug window (and so is our settings window). So it would essentially mean having independent windows open at the same time which may be not trivial (multithreading).

craigmaloney:
Right now for me (under Ubuntu Linux) whenever I hit F8 to go into the monitor it will close the "CRT" screen. I'm left with the debugger/monitor in a text window. Which is fine, but I would like to have the CRT screen left up if possible. Additionally, having it display the scan line and the CRT position that it's currently processing in some way would be extremely handy, 

Thanks for considering this.

groessler:
On 7/17/23 21:30, Craig Maloney wrote:
>
> Right now for me (under Ubuntu Linux) whenever I hit F8 to go into the 
> monitor it will close the "CRT" screen. I'm left with the 
> debugger/monitor in a text window. Which is fine, but I would like to 
> have the CRT screen left up if possible. Additionally, having it 
> display the scan line and the CRT position that it's currently 
> processing in some way would be extremely handy,
>

The X11 target doesn't close the emulation window when entering the 
debugger. Very nice for debugging.


regards,
chris

craigmaloney:
I've done some testing under Linux with the following patch.

```
diff --git a/src/sdl/main.c b/src/sdl/main.c
index 228fbad6..ae39f2fe 100644
--- a/src/sdl/main.c
+++ b/src/sdl/main.c
@@ -108,7 +108,7 @@ int PLATFORM_Exit(int run_monitor)
           for the time the monitor is active. Also, with SDL_VIDEODRIVER=directx all
[...]

PR #213, opened 2024-01-03
Close SDL audio subsystem when entering monitor

### LINKED ISSUE #97: High CPU usage in monitor?
I run atari800 (v. 3.1.0-r1 on Gentoo Linux), then go to the Monitor by hitting F8, the simulation is **paused** and one might expect that it should not grind the CPU too much, right? :q 

And yet, after a couple of minutes, it **starts using 70..100% CPU** despite that I'm doing nothing in the Monitor and the simulation is paused (the simulation window disappeared).

I attached GDB to it to check what function does it run. There seems to be three threads, one of them being the main thread. The main thread doesn't seem to do much, just waiting for user input in `readline` routine:

  main → Atari800_Frame → UI_Run → Atari800_Exit → PLATFORM_Exit → MONITOR_Run → safe_gets → {1 readline → readline_internal_char→ rl_read_key → rl_getc → {0 pselect } }

(I marked library boundaries with numbered braces: 0=libc, 1=libreadline, 2=libpthread, 3=libSDL-1.2, 4=libpulse, 5=libpulsecommon, 6=libasound, 7=libasound_module_pcm_pulse.)

The other two threads call some stuff from `libSDL` and `PulseAudio`:

{0 clone → {2 start_thread → {5 ? → {4 ? → pa_mainloop_run → pa_mainloop_iterate → pa_mainloop_poll → {0 poll} } }

{0 clone → {2 start_thread → {3 ? → ? → ? → ? → {6 ? → ? → ? → ? → {7 ? → ? → {4 pa_stream_writable_size → {5 pa_detect_fork → {0 getpid} } } } } } } }
so my guess is that there is a problem with some callbacks from the sound API racing with each other. Another hint that this might be the case is that I sometimes get the following message on the terminal:

ALSA lib /{...redacted...}/alsa-lib-1.1.2/src/pcm/pcm.c:8251:(snd_pcm_recover) underrun occurred

and the CPU usage seems to raise a while after the first of these messages appear.

Ahbne33:
On some Linux systems merely pausing SDL audio while SDL video is disabled, causes CPU usage to jump to 100% after a few minutes. Avoid the issue by closing SDL audio when entering the monitor, and reopening it when exiting it.

Fixes #97.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -97,7 +97,7 @@ atari800_SOURCES += \
 	sndsave.c sndsave.h
 endif
 if WITH_SOUND_SDL
-atari800_SOURCES += sound.c sound.h sdl/sound.c
+atari800_SOURCES += sound.c sound.h sdl/sound.c sdl/sound.h
 endif
 if WITH_SOUND_FALCON
 atari800_SOURCES += sound.c falcon/sound.c
--- a/src/sdl/main.c
+++ b/src/sdl/main.c
@@ -43,6 +43,7 @@
 #include "platform.h"
 #ifdef SOUND
 #include "../sound.h"
+#include "sdl/sound.h"
 #endif
 #ifdef USE_UI_BASIC_ONSCREEN_KEYBOARD
 #include "akey.h"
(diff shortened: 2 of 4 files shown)


craigmaloney:
Wouldn't it be better to leave the SDL video up? #188 is the issue that I'd like to see addressed where the screen stays on.

joysfera:
Yea, let's wait with this and better try looking into #188

mikrosk:
Not sure why @Ahbne33 didn't mention it but this is basically just @kr0tki's branch from #97. So closing as duplicate (and reopened #97 as I mistakenly thought that his commit already made it to master).

And agreed, let's investigate this in context of #188 before doing further steps.

PR #221, opened 2024-01-27
Add SDL2 build option

mikekov:
This is a WIP to add SDL2 support. Basic functionality is here, but it's been only tested on macOS. To try it out:

`./configure --with-video=sdl2 --with-sound=sdl2`

or, on macOS with homebrew:

`./configure --with-video=sdl2 --with-sound=sdl2 CFLAGS="-I/opt/homebrew/include -g"`

I need some help with `configure.ac:1389`, as bash conditionals are dense.

OpenGL >= 3.3 support is TODO. Currently 2.1 is used.


mikrosk:
Thanks for this initiative! I've briefly tried it:
- it compiles and runs on Linux
- however keyboard doesn't seem to work at all
- default `./configure` is not compilable anymore:
```
../../atari800.git/src/videomode.c: In function ‘UpdateVideoWindowed’:
../../atari800.git/src/videomode.c:593:9: error: C++ style comments are not allowed in ISO C90
  593 |         //TODO:
      |         ^
../../atari800.git/src/videomode.c:593:9: note: (this will be reported only once per input file)
make[2]: [Makefile:2940: atari800-videomode.o] Error 1 (ignored)
../../atari800.git/src/sdl/input.c: In function ‘PLATFORM_GetJoystickKeyName’:
../../atari800.git/src/sdl/input.c:374:17: error: unknown type name ‘SDL_Keycode’
  374 |         typedef SDL_Keycode SDLKey;
```

> I need some help with configure.ac:1389, as bash conditionals are dense.

What exactly you would like to solve there?

Maybe you could install a VM on your Mac for easier debugging / bug hunting?

mikekov:
I need this line in `configure.ac`:
`if [[ "$a8_target" != default -o "$with_video" != sdl2 ]]; then`
to allow *both* "sdl" and "sdl2" targets. Right now I hijacked it to test for "sdl2" only.

It needs to check this condition: "if a8_target != default AND with_video != sdl AND with_video != sdl2".

mikrosk:
Sorry, I'm still confused. You're struggling with that exactly? Replacing `-o` with `-a` or with adding a new expression or there's missing some video definition or ... ?

mikekov:
I don't know the syntax of *.ac files at all; I don't use automake tools, so I have no clue what the priorities of operators in those conditionals are.

As for keyboard not working, I'd try with a fresh ~/.atari800.cfg. SDL2 has redefined key and keyscan values.

sba1:
@ mikekov 
You may also have a look look at https://github.com/sba1/atari800/tree/sdl2 where I also started to conversion to SDL2 few years ago. Maybe it is useful.

mikekov:
Some video effects are now implemented by shaders. Here's CRT monitor barrel distortion:

atari-barrel

CRT-like pixel shape, from blocky (no effect) to rounded:

atari-beam

...and phosphor glow:

atari-glow

Scanline effect is also moved to a shader.

mikrosk:
Cool stuff!

joysfera:
Pretty wild, TBH. Looks better than original :-)

mikekov:
Thank you guys. Fixing controller issues now. It didn't work in SDL2 out of the box, but once I overcame that, it turns out analog controller doesn't work very well with existing axes decoding. Some games that ignore diagonal joystick positions suffer badly and are hardly playable. This is because they treat diagonal positions as either left or right, which makes pointing up/down harder than it should be. I'm thinking about adding some options to joystick settings to change how diagonals are reported:
image
Some games, like Ninja, rely on joystick diagonal positions to unlock full range of player moves and it is beneficial to have diagonals easily achievable (that's wide diagonal zones). Some, like Boulder Dash, are best played with no diagonals decoded/reported. There could perhaps be middle ground (closer to how digital joystick feels), so narrow diagonal zones can be used too.
Let me know what you think. Arguably, this falls into controller per-game configuration, which complicates things further. At the moment I'd limit the scope of changes by making those into "real joystick" settings.

joysfera:
We're not going into a per-game configuration.
I wouldn't spend too much time on analog joysticks because we have either joypads or true digital joysticks if you need diagonals so much. 
From your images above the existing decoding seems OK to me. If not the narrow diagonals would be also acceptable, IMHO. The wide diagonals could generate too many "false positives" that would upset users.

IMHO the analog joysticks should me mapped to paddles almost exclusively. For directional joysticks we should use either joypads or true digital joysticks. Analog joysticks are sort of last resort here, I think.

mikekov:
Thank you for the feedback, I'll continue with minimal configuration approach enhancing real joysticks setup.

I made D-pad work in SDL2 (it uses button query in SDL2) and it functions fine now. Analog sticks a just a bonus, and they work well too. I tried connecting them to a paddle input, but they are too sensitive for that. Plus, different games use different effective range of paddle values, so that implementation can wait for other time.

As for the existing analog joystick decoding, here's where it falls flat:
image
In BD, up/left & down/left are interpreted as simply left direction. That's not really usable, so this is where new config option will help.

mikekov:
atari-joystick

joysfera:
The enhanced joystick configuration looks good. Is it implemented for SDL1 as well?
Also, there should be a reasonable default for the Diagonals zone and I don't think the "None" is it. The "Narrow" choice as the default seems better to me.

mikekov:
SDL1 implementation is left completely intact. I didn't want to upset anything there.

As for a default diagonals zone it is actually "Narrow" one. I tested it by deleting `~/atari800.cfg` first.

mikekov:
* Added ability to navigate UI menu with D-pad buttons
* Added controller button configuration

atari-joystick-btn-config

joysfera:
I feel like you're committing too many changes under one pull request. Quick save state does not sound like related to "SDL2 build option". It would be better to implement new features separately from SDL2.

mikekov:
I got carried away by controller configuration options. I think first and third commit only need to be cherry-picked for SDL2 work. I'll create separate PRs.

PR #222, opened 2024-01-29
added sumaries and error codes to some monitor commands; extended mon…

jhusak:
…itor fill command to fill with many bytes; also added length read parameter optional.


jhusak:
Ok.

PR #224, opened 2024-02-09
SDL2 support

mikekov:
This PR attempts to add SDL2 build support. It builds on top of existing SDL1 code, modifying conditionally all the places where SDL2 functions syntax is different.
OpenGL implementation adds support for running vertex and shader programs. They are included in a source form (atari800-shader.*), as well as pre-generated (with xxd script, also checked in) to allow for inclusion in the C source code.

To use SDL2: `./configure --with-video=sdl2 --with-sound=sdl2`

---

Note: `configure.ac:1389` needs attention


mikrosk:
Thanks for splitting #221. Tested on Arch Linux (usually latest versions of everything). See my notes below.

Make sure you don't commit trailing spaces  (just saw one at line 3891 in ui.c).

The block in configure.ac which you're looking for to fix is:
```
if [[ "$a8_target" != default -o \( "$with_video" != sdl -a "$with_video" != sdl2 \) ]]; then
    if [[ "$with_opengl" = "yes" ]]; then
        AC_MSG_ERROR([option --with-opengl only supported with --target=default and with --with-video=sdl or --with-video=sdl2!])
    fi
```
There are some warnings by `./configure && make`:
```
../../src/sdl/input.c: In function ‘PLATFORM_Keyboard’:
../../src/sdl/input.c:687:23: warning: missing terminating ' character
  687 |         //TODO - there's no mouse grab function in SDL2
      |                       ^
```
```
../../src/ui.c:2484:13: warning: ‘CrtGlowSliderLabel’ defined but not used [-Wunused-function]
 2484 | static void CrtGlowSliderLabel(char *label, int value, void *user_data) {
      |             ^~~~~~~~~~~~~~~~~~
../../src/ui.c:2479:13: warning: ‘CrtBeamSliderLabel’ defined but not used [-Wunused-function]
 2479 | static void CrtBeamSliderLabel(char *label, int value, void *user_data) {
      |             ^~~~~~~~~~~~~~~~~~
../../src/ui.c:2474:13: warning: ‘CrtBarrelSliderLabel’ defined but not used [-Wunused-function]
 2474 | static void CrtBarrelSliderLabel(char *label, int value, void *user_data) {
```
There's an error by `./configure --with-video=sdl2 --with-sound=sdl2 && make`:
```
../../src/sdl/input.c: In function ‘get_SDL_joystick_state’:
../../src/sdl/input.c:1681:59: error: ‘M_PI’ undeclared (first use in this function)
 1681 |         double angle = (atan2(x / 32768.0, y / 32768.0) + M_PI) * 180 / M_PI;
      |                                                           ^~~~
../../src/sdl/input.c:1681:59: note: each undeclared identifier is reported only once for each function it appears in
```
```
../../src/sdl/video_gl.c: In function ‘InitGlFunctions’:
../../src/sdl/video_gl.c:533:9: error: expected expression before ‘#pragma’
  533 | #pragma GCC diagnostic push
      |         ^~~
../../src/sdl/video_gl.c: In function ‘SetOrtho’:
../../src/sdl/video_gl.c:830:25: error: ‘M_PI’ undeclared (first use in this function)
  830 |         angle = angle * M_PI / 180;
      |                         ^~~~
../../src/sdl/video_gl.c:830:25: note: each undeclared identifier is reported only once for each function it appears in
../../src/sdl/video_gl.c: At top level:
../../src/sdl/video_gl.c:839:14: warning: ‘read_text_file’ defined but not used [-Wunused-function]
  839 | static char* read_text_file(const char* path) {
      |              ^~~~~~~~~~~~~~
```

And most importantly, as  mentioned in #221, keyboard input doesn't work at all. 

Apart from that, your changes look good. If you manage to fix the issues above within a few days and testing will pass afterwards, we can still aim to including it into the upcoming release. :)

mikekov:
Thanks for review and makefile hints Miro. I hope to eventually fix the issues, but I don't know about squeezing it in the next few days, I'm not that fast ;-) Especially since keyboard handling seems to be "borken" on Linux. Maybe next next release...

mikekov:
OK, I have just built this PR on KdeNeon and keyboard works fine. But I suspect the nonworking one is due to the SDL2 redefined key codes and existing `.atari800.cfg` clashing with them. If you delete it or move it aside, then keyboard should be functional.

mikrosk:
Hmm, I can't rule that one out, I'll test tonight and let you know.

However backward compatibility would be nice, too. Some user cherish their configuration files for years, so we can't just ask them to delete it.

mikrosk:
Oh and please rebase your changes on master, there are some conflicts now, perhaps due to `win32` removal.

mikekov:
OK, I fixed warnings and rebased my changes. I changed key binding names used in `.atari800.cfg`, so they won't clash with SDL1 settings.

mikrosk:
SDL1 build now seems to be completely undisturbed and what's better, in SDL2 my keyboard works again, so yes, the old key bindings were the culprit.

First the usual set of compiler complains during `./configure --with-video=sdl2 --with-sound=sdl2 && make`:
```
../../atari800.git/src/ui.c: In function ‘TapeSliderLabel’:
../../atari800.git/src/ui.c:1525:40: warning: ‘snprintf’ output may be truncated before the last format character [-Wformat-truncation=]
 1525 |                 snprintf(label, 10, "%u", (unsigned int)value + 1);
      |                                        ^
../../atari800.git/src/ui.c:1525:17: note: ‘snprintf’ output between 2 and 11 bytes into a destination of size 10
 1525 |                 snprintf(label, 10, "%u", (unsigned int)value + 1);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
...
```
../../atari800.git/src/sdl/video_gl.c:849:14: warning: ‘read_text_file’ defined but not used [-Wunused-function]
  849 | static char* read_text_file(const char* path) {
      |              ^~~~~~~~~~~~~~
../../atari800.git/src/sdl/video_gl.c: In function ‘SDL_VIDEO_GL_SetVideoMode’:
../../atari800.git/src/sdl/video_gl.c:931:25: warning: ‘strncpy’ output truncated before terminating nul copying 185 bytes from a string of the same length [-Wstringop-truncation]
  931 |                         strncpy(vertexShader, (char*)vertexShaderArr, vertexShaderArr_len);
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../atari800.git/src/sdl/video_gl.c:935:25: warning: ‘strncpy’ output truncated before terminating nul copying 5406 bytes from a string of the same length [-Wstringop-truncation]
  935 |                         strncpy(fragmentShader, (char*)fragmentShaderArr, fragmentShaderArr_len);
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Then I have noticed that those are not remapped from SDL1:
```
Unrecognized variable or bad parameters: 'SDL_JOY_0_ENABLED=1'
Unrecognized variable or bad parameters: 'SDL_JOY_0_LEFT=260'
Unrecognized variable or bad parameters: 'SDL_JOY_0_RIGHT=262'
Unrecognized variable or bad parameters: 'SDL_JOY_0_UP=264'
Unrecognized variable or bad parameters: 'SDL_JOY_0_DOWN=261'
Unrecognized variable or bad parameters: 'SDL_JOY_0_TRIGGER=305'
Unrecognized variable or bad parameters: 'SDL_JOY_1_ENABLED=0'
Unrecognized variable or bad parameters: 'SDL_JOY_1_LEFT=97'
Unrecognized variable or bad parameters: 'SDL_JOY_1_RIGHT=100'
Unrecognized variable or bad parameters: 'SDL_JOY_1_UP=119'
Unrecognized variable or bad parameters: 'SDL_JOY_1_DOWN=115'
Unrecognized variable or bad parameters: 'SDL_JOY_1_TRIGGER=306'
Unrecognized variable or bad parameters: 'SDL_UI_KEY=282'
Unrecognized variable or bad parameters: 'SDL_OPTION_KEY=283'
Unrecognized variable or bad parameters: 'SDL_SELECT_KEY=284'
Unrecognized variable or bad parameters: 'SDL_START_KEY=285'
Unrecognized variable or bad parameters: 'SDL_RESET_KEY=286'
Unrecognized variable or bad parameters: 'SDL_HELP_KEY=287'
Unrecognized variable or bad parameters: 'SDL_BREAK_KEY=288'
Unrecognized variable or bad parameters: 'SDL_MON_KEY=289'
Unrecognized variable or bad parameters: 'SDL_EXIT_KEY=290'
Unrecognized variable or bad parameters: 'SDL_SSHOT_KEY=291'
Unrecognized variable or bad parameters: 'SDL_TURBO_KEY=293'
Unrecognized variable or bad parameters: 'SDL_JOY_0_USE_HAT=0'
Unrecognized variable or bad parameters: 'SDL_JOY_1_USE_HAT=0'
Unrecognized variable or bad parameters: 'SDL_JOY_2_USE_HAT=0'
Unrecognized variable or bad parameters: 'SDL_JOY_3_USE_HAT=0'
```
I'm sure we can do something about at least some of them?

Then I could finally see your changes in action, pretty cool stuff! Few nitpicks:
- I had to enable Hardware Acceleration in the video menu to make any of the new options to take affect. If SDL2 requires hardware acceleration, we should perhaps just enable it and remove this option from the menu. If iSDL works fine without HW acceleration (which it seems to be...) perhaps we should enable it by default for SDL2 anyway and if you disable it, also hide the three CRT options?
- When I did enable it, the effects worked however I got white background around the menu (perhaps an effect from a window resize event):
image
- why bilinear filtering and pixel buffer objects (what is it?) have to be disabled in SDL2?

Apart from those, I would see your code is good to merge.

mikekov:
Thanks for testing.
* SDL2 supports both hardware (OpenGL) and software rendering. Only OpenGL version supports shaders and CRT effects.
* You are right, I think hardware support should be on by default on SDL2 and CRT options hidden or disabled on software renderer.
* Bilinear filtering is disabled (hidden) because it was used to simulate soft TV output, and this is now done though the shaders, which offer much more control over display rendering.
* Pixel buffer objects are disabled simply because OpenGL code path in SDL2 does not use them. I'm not an expert here, I don't know what they were used for in SDL1 implementation.
* I'm not sure what causes white border in your output; this is wrong and it was supposed to show simulated gray bezel.
* `SDL_JOY_0_ENABLED` and the like are not remapped, because SDL2 redefined completely all key codes and key scans, so they are not interchangeable. It is unfortunately necessary to press physical keys to map them again.

mikrosk:
@mikekov any chance you could clean up / implement the points from https://github.com/atari800/atari800/pull/224#issuecomment-2081852452 so we can merge this one?

Same for #221, if you could provide the new joystick mapping as a separate PR (or anything else which could be considered as a separate feature) so we can merge that, too.

I was talking to @joysfera, we would like to include your code in the next release (although maybe disabled by default) so users are free to experiment with the new SDL2 frontend.

mikekov:
I rebased and resolved the conflicts. CRT video options are now only presented when OpenGL is on. I'm using clang and I don't see warnings you reported.

mikrosk:
OK, I have finally found some time to look into this again. This is basically clean to merge, except the last three items:

- hardware acceleration is still OFF by default in SDL2 builds
- the white border is also still present... I'm attaching my config file => that way you can test that when HW acceleration is enabled + CRT barrel distortion is changed, the white background appears
- warnings are also still there (as posted above); I think you don't need `gcc` to see what needs to be fixed

[atari800.cfg.zip](https://github.com/user-attachments/files/17447324/atari800.cfg.zip)

@mikekov if you find some time, please try to solve them and we are free to go!

mikrosk:
@mikekov any  chance you could look into the issues above? We plan to make a new release soon(-ish) and we'd like to include your work there.

This applies also to https://github.com/atari800/atari800/pull/221, that one needs to be rebased on this one (i.e. the PR would offer the new joystick functionality alone).

mikekov:
I have mostly forgotten what I've been doing here and lost the context, it's been over a year. Frankly, at this point I'd prefer if somebody else picked up where I left off. This port to sdl2 should be mostly complete.

yuppox:
SDL1.2 was deprecated a long time ago and hasn't had a release in over 10 years. I understand it's needed for certain platforms but SDL2 should be an option.

I was able to get the SDL2 fork to compile but there was no keyboard input. I'm on embedded Linux (Armbian).

I also tried SDL12-compat. It worked but I don't think the performance was especially good. One CPU core stays at 96-99%, which seems excessive to me. It's using OpenGL, so I'd expect less CPU usage.

mikekov:
Keyboard should work. Maybe its definition clashes with sdl1. Try renaming preferences, so they can be recreated. I don't remember if I found a workaround. sdl2 changed all key definitions.

PR #230, opened 2024-04-22
J(atari)Carts & DCart

GienekP:
Added carts based on "Atarimax 1 MB Flash cartridge (new)" but witch different sizes.


DOC/cart.txt
```diff
@@ -1247,7 +1256,7 @@ Reference: http://www.retrobits.net/atari/osshack.shtml
 | Type 77: MIO diagnostics 8KB cartridge                                    |
 +---------------------------------------------------------------------------+
 
- The cartridge uses the same board as type 11 - SpartaDOS X, but uses only
+ The cartridge uses the same board as type 11 - Sparta Dos X, but uses only
```
mikrosk, line 1259: Why this change? SpartaDOS X was officially stylized in such way (see e.g. https://en.wikipedia.org/wiki/SpartaDOS_X), so I don't think this change is justified.

GienekP: > Why this change? SpartaDOS X was officially stylized in such way (see e.g. https://en.wikipedia.org/wiki/SpartaDOS_X), so I don't think this change is justified.

You're right. I just used the previous version of "cart.txt".

mikrosk: If that's the case, can I ask you to change it also on other places so it says "SpartaDOS X" everywhere?

Apart from that, your PR is very much welcome, thanks!

mikrosk:
Merged with slight formatting changes.  Thanks!

PR #231, opened 2024-04-25
Another go at fixing #79

### LINKED ISSUE #79: DLI timing problem
At AtariAge, someone mentioned, that the demo "Second Life Syndrome" (xex-file zipped and attached) is not running on Atari800MacX and atari800 (which is the base for Atari800MacX).
The demo runs fine on real hardware and on the Altirra-Emulator by phaeron (Windows only).

phaeron seems to have found the issue:

"Under normal circumstances, a DLI handler would try to begin executing at cycle 10, two cycles after the end of the LMS fetches, whenever the next instruction border occurs. However, if a taken branch is right before that, it delays interrupt acknowledgment and in this case it allows the DEC $1827 instruction that dmsc noted above to sneak in. This instruction is critical in this case because (1) the DLI routine is too long and keeps interrupting itself and (2) the DLI is also responsible for revectoring the NMI vector to the VBI handler based on this counter. With that instruction being skipped, the DLI code doesn't ever branch to the code to set up the VBI. The CPU does eventually run all of the decrements when the stack unwinds at the end of the frame when the DLIs stop, but all of the decrements run back-to-back without the checks in between, so the X=$00 condition is never seen by the branch."

Find the full topic there: https://atariage.com/forums/topic/300257-second-life-demo-lamers-on-atari800mac-501/
[LMS.XEX.zip](https://github.com/atari800/atari800/files/4023305/LMS.XEX.zip)

dmsc:
This is simpler than the old fix and should better take into account the real NMI delay on branches.
Fixes #79.

### DIFF
--- a/src/antic.c
+++ b/src/antic.c
@@ -429,13 +429,13 @@ unsigned int ANTIC_screenline_cpu_clock = 0;
 		draw_antic_ptr = draw_antic_table[GTIA_PRIOR >> 6][anticmode];\
 		gtia_bug_active = FALSE;\
 	}}while(0)
-#define GOEOL_CYCLE_EXACT  CPU_GO(ANTIC_antic2cpu_ptr[ANTIC_LINE_C], FALSE); \
+#define GOEOL_CYCLE_EXACT  CPU_GO(ANTIC_antic2cpu_ptr[ANTIC_LINE_C]); \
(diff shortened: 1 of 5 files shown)


mikrosk:
Fixed with a different patch (but still kept the credit of course).

PR #232, opened 2024-05-12
Adds github workflow to build on Ubuntu 20.04

dmsc:
This adds a github workflow that builds the code on Ubuntu 20.04 for 64-bit x86 and uploads the resulting binary.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,29 @@
+name: Build CI
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ "master" ]
+  pull_request:
+
+jobs:
+  build-amd64:
+    name: Build on Linux amd64
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: install depends
+      run: sudo apt-get install -y libsdl1.2-dev libsdl2-dev libmp3lame-dev
+    - name: autogen
+      run: ./autogen.sh
+    - name: configure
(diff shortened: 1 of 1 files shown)


joysfera:
I am not sure if we can navigate people to the artifacts for downloading the binaries - the last stable URL is to /actions/. Also no idea how many artifacts can you left stored on github and when/if they purge it eventually.

I can provide a space on one of my servers for uploading the binaries onto a stable URL (pity I let go the atari800 domain, now it would be handy). 

Maybe we could store few latest binaries in the atari800.github.io space unless it's against a github rule. If it is then see above for my offer of a server space.

dmsc:
Hi!

> I am not sure if we can navigate people to the artifacts for downloading the binaries - the last stable URL is to /actions/. Also no idea how many artifacts can you left stored on github and when/if they purge it eventually.
> 

The idea is that you can download the artifact and attach to a bug-report, to let the user test specific versions. Adding MacOS and Windows support (it is easy after this is merged) will make testing easier, IMHO.

The artifacts are stored in Github for 90 days, so it is plenty of time for testig.

Have Fun!

joysfera:
Definitely cool idea, but I am also thinking about "Nightly builds" (or basically builds after every commit) that would be stored somewhere kinda forever so the Atari800 web could contain links to "latest builds".

mikrosk:
@dmsc great stuff, I was completely unaware of this possibility.

@joysfera one way to do it, without external servers, is basically to create a Github Action for publishing a pre-release and updating that pre-release on every commit. That way you get a stable url and very visible download location.

PR #233, opened 2024-05-13
Adds MacOS 14 in arm64 to CI workflow.

dmsc:
Adds MacOS to the CI.


dmsc:
Hi!

You can see now at https://github.com/atari800/atari800/actions/runs/9066780948?pr=233 the download for both Linux and MacOS versions:

image

PR #248, opened 2025-06-27
don't fread 0 bytes when length of the FUJI chunk is 0

mono6502:
Ninja from Mastertronic from Atarimania https://www.atarimania.com/game-atari-400-800-xl-xe-ninja_3714.html doesn't want to be loaded. Short investigation appeared that this file hasn't got description and FUJI chunk contains length field set to 0, but atari800 calls fread with 0 bytes. Bugfix makes a call only when length is greater than 0 and Ninja works fine.


mikrosk:
Nice and simple, thanks!

PR #249, opened 2025-06-27
Cas no description

mono6502:
I forgot to commit afile.c change with exchanged an order of .CAS and .XFD detection.


mikrosk:
Merged with a slightly updated description, thanks!

PR #253, opened 2025-11-08
fix compilation break if project configured without opengl

marcin-sochacki:
SDL_VIDEO_opengl is not declared if the project is configured with:
--with-opengl=no


joysfera:
Thanks!

PR #264, opened 2026-03-18
netsio: implement PROCEED signal handling

ericpareja:
the following patch implements PROCEED signal handling in netsio for communication with fujinet.


mikrosk:
Looks good to me. Btw, it nicely demonstrates why "pre-writing" some code for future use isn't very beneficial -- the re-activated functions differ quite a bit in their final form.

PR #265, opened 2026-03-22
netsio: add startup wait and gate netstream mode via Fuji enable + motor + POKEY config

mozzwald:
Adds FujiNet Netstream state machine/hooks in netsio/pokey/sio/pia which allows using the SIO port as a raw serial to network device.

Adds a brief wait to emulation startup for NetSIO connection to be made, otherwise you may need to perform cold reset before emulator boots from FujiNet.

### DIFF
--- a/src/atari.c
+++ b/src/atari.c
@@ -149,6 +149,9 @@
 #endif
 #ifdef NETSIO
 #include "netsio.h"
+
+#define NETSIO_STARTUP_WAIT_TIMEOUT_MS 5000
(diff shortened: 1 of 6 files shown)


mikrosk:
Merged. Btw @mozzwald would you like to have granted push access to the repository? You are the most qualified one for judging quality of NetSIO commits. If you need to touch other parts of code, you can always create a PR for others to see but otherwise I'm fine with you changing netsio files as you need.

Issue #206, opened 2023-12-02
Keyboard issues on Falcon


mikrosk:
While testing #205 I have noticed:
- ~~Atari SDL doesn't propagate SHIFT and CAPS (they are just ignored), however CONTROL works thanks to #205~~
- Atari Falcon doesn't propagate even CONTROL, CONTROL+A-Z produces strange characters (like %, ", ...)
- ~~most likely https://github.com/atari800/atari800/pull/203 needs to me adapted as well~~

PR #266, opened 2026-04-02
Fix control-letter key handling in SDL2 to match SDL1.2

ericpareja:
Key handling of control-letter combinations were not properly handled when compiling with SDL2, but were working in SDL1.2.  Changes in input and key press handling from SDL1.2 to SDL2.0 resulted in a difference in returned values for keycodes/scancodes. The value of lastuni was not being passed properly when in SDL2.0. This patch fixes that.

Test compiled:  --with-video=sdl and --with-video=sdl2
Manually tested both builds to determine if keypress behavior for control-key combinations in both builds matched.

### DIFF
--- a/src/sdl/input.c
+++ b/src/sdl/input.c
@@ -1324,6 +1324,15 @@ int PLATFORM_Keyboard(void)
 	if (key_control) {
 		if (lastuni == 0 && lastkey >= SDLK_a && lastkey <= SDLK_z)
 			lastuni = lastkey - SDLK_a + 1;
+		/* lastuni is set 0 only in SDL2 even when Ctrl key
+		   is pressed along with another key but we know that
+		   the control key is currently down so we set lastuni
+		   to the ASCII equivalent for control code (1-26)
+		   only for alpha keys so that they can be handled
+		   properly below.
+		   in SDL1.2, lastuni is set properly already, so it
+		   will have a non-zero value.
+		*/
 		switch(lastuni) {
 		case '.':
 			return AKEY_FULLSTOP|shiftctrl;
--- a/src/sdl/input.c
+++ b/src/sdl/input.c
@@ -1322,6 +1322,8 @@ int PLATFORM_Keyboard(void)
 
 	/* Handle CTRL-0 to CTRL-9 and other control characters */
 	if (key_control) {
+		if (lastuni == 0 && lastkey >= SDLK_a && lastkey <= SDLK_z)
+			lastuni = lastkey - SDLK_a + 1;
 		switch(lastuni) {
 		case '.':
 			return AKEY_FULLSTOP|shiftctrl;


ericpareja:
The Alt-Ctrl-J toggle keyboard joysticks on/off patch was sent to me by B. Watson (Urchlay) which I modified to work with SDL2 as well as SDL1.2.

ericpareja:
Related to #205 and #206

mikrosk:
@ericpareja a couple of notes:

- commit https://github.com/atari800/atari800/pull/266/changes/5a62ff743561b9c4044900a6e9ec98ac537ac934 could use some explanation what is it doing and why (esp. in comparison to SDL 1.2; maybe explain in a greater detail how it is related to #205 and #206?)
- commit https://github.com/atari800/atari800/pull/266/changes/6eee0bed4ef466b666e4c0aae313a83edb8b10e1 mentions CTRL and ALT but I see only CTRL being tested?

ericpareja:
> @ericpareja a couple of notes:
> 
>     * commit [6eee0be](https://github.com/atari800/atari800/commit/6eee0bed4ef466b666e4c0aae313a83edb8b10e1) mentions CTRL and ALT but I see only CTRL being tested?

In line 862, if you expand upward, you'll see:

`if (get_key_state(kbhits, SDLK_LALT)) {`

which tests for Left Alt key before proceeding to the CTRL key being tested.

ericpareja:
> @ericpareja a couple of notes:
> 
>     * commit [5a62ff7](https://github.com/atari800/atari800/commit/5a62ff743561b9c4044900a6e9ec98ac537ac934) could use some explanation what is it doing and why (esp. in comparison to SDL 1.2;

The code section starting with line 1323 is for Handling control-character combinations
```C
/* Handle CTRL-0 to CTRL-9 and other control characters */
	if (key_control) {
		switch(lastuni) {
```
Note that the switch is controlled by `lastuni`.

In lines ~664-668, lastuni is defined by:
```C
#if SDL2
                        lastuni = 0;
#else
                         lastuni = event.key.keysym.unicode;
#endif
```
Note that lastuni is 0 for SDL2.
If ctrl-key was already detected, the problem now is lastuni is 0, and this will fall through to 
line 1609
```
return AKEY_NONE;
```
 which isn't what we want. We want `lastuni` to have a value similar to SDL1.2 so that one of the conditions in lines 1375-onwards is triggered.

Line 1325-1326 in our fix:
```
		if (lastuni == 0 && lastkey >= SDLK_a && lastkey <= SDLK_z)
			lastuni = lastkey - SDLK_a + 1;
```
If we're in SDL1.2, lastuni will be non-zero. lastuni is zero only in SDL2, so there's no need for a guard #ifdef SDL2 here. The expression `lastuni = lastkey - SDLK_a + 1` sets lastuni to a corresponding ASCII control code (1-26), which is triggered later in lines 1375-onwards, just like in SDL1.2.

> maybe explain in a greater detail how it is related to [Fixes CTRL and CAPS keys when using SDL12-compat library. #205](https://github.com/atari800/atari800/pull/205) and [Keyboard issues on Falcon #206](https://github.com/atari800/atari800/issues/206)?)

I submitted another proposed fix to handling CAPS LOCK, because all these fixes are related to changes in keyboard handling in SDL1.2 to SDL2.

joysfera:
Thanks for fixing the Ctrl hotkey (I have my own hack but haven't got to merging it yet) and for the Caps Lock clean up. I'd merge both immediately.

As for the hotkey for the joystick I understand why it is useful/needed but I am not sure if I like that.

ericpareja:
> Thanks for fixing the Ctrl hotkey (I have my own hack but haven't got to merging it yet) and for the Caps Lock clean up. I'd merge both immediately.
> 
> As for the hotkey for the joystick I understand why it is useful/needed but I am not sure if I like that.

The hotkey for the keyboard joystick toggle is simply a modified hotkey for joystick swapping. Alt-J (swap joysticks) vs Ctrl-Alt-J (keyboard joystick toggle).

mikrosk:
@ericpareja 

> If we're in SDL1.2, lastuni will be non-zero. lastuni is zero only in SDL2, so there's no need for a guard #ifdef SDL2 here. The expression lastuni = lastkey - SDLK_a + 1 sets lastuni to a corresponding ASCII control code (1-26), which is triggered later in lines 1375-onwards, just like in SDL1.2.

Please mention this explicitly in the code that you rely on `lastuni` being set to zero by SDL2 (if it is not possible to set it earlier, similar to SDL1.2?)

> I submitted another proposed fix to handling CAPS LOCK, because all these fixes are related to changes in keyboard handling in SDL1.2 to SDL2.

Carefully with removing that hack -- there are basically three cases to handle (see https://github.com/libsdl-org/sdl12-compat/issues/331):

- original SDL1.2 behaviour
- SDL1.2-compat behaviour before 1270 version
- SDL1.2-compat behaviour after 1270 version
- (and of course SDL2 behaviour)

So you should build atari800 against all three of them (i.e. you need to install an older/newer sdl1.2 and sdl1.2-compat somewhere and link against that library) to claim this being fixed / improved. Otherwise it would re-introduce the same regressions.

joysfera:
What if we kinda dropped (or maybe just abandoned) support for SDL < 2 given that current version is 3.4.4 already? Or is it too early? Frankly I'd love to upgrade to SDL 3 ASAP because it supports certain platforms previous versions don't.

joysfera:
> The hotkey for the keyboard joystick toggle is simply a modified hotkey for joystick swapping. Alt-J (swap joysticks) vs Ctrl-Alt-J (keyboard joystick toggle).

I know what it does. But I feel like the whole joystick framework needs improving (and I've been "working" on that for more than a year now). For example I think the idea of SwapJoysticks() is no longer useful (if it ever was). It makes more sense to me to use that hotkey just for enabling or disabling the keyboard joysticks (all of them at once). At the same time the keyboard joysticks should have some kind of visual indicator (similar to disk drive activity). But that's for another github issue/pull request apparently.

ericpareja:
> > The hotkey for the keyboard joystick toggle is simply a modified hotkey for joystick swapping. Alt-J (swap joysticks) vs Ctrl-Alt-J (keyboard joystick toggle).
> 
> I know what it does. But I feel like the whole joystick framework needs improving (and I've been "working" on that for more than a year now). For example I think the idea of SwapJoysticks() is no longer useful (if it ever was). It makes more sense to me to use that hotkey just for enabling or disabling the keyboard joysticks (all of them at once). At the same time the keyboard joysticks should have some kind of visual indicator (similar to disk drive activity). But that's for another github issue/pull request apparently.

I will separate the keyboard joystick patch from the SDL1.2/SDL2 fixes then and maybe submit a separate PR or open an issue for it.

joysfera:
> I will separate the keyboard joystick patch from the SDL1.2/SDL2 fixes then and maybe submit a separate PR or open an issue for it.

Thank you! It'll allow us to merge the much needed Ctrl fix ASAP.

mikrosk:
> What if we kinda dropped (or maybe just abandoned) support for SDL < 2 given that current version is 3.4.4 already? Or is it too early? Frankly I'd love to upgrade to SDL 3 ASAP because it supports certain platforms previous versions don't.

Nothing prevents adding SDL3 support (or better, replacing SDL2 with SDL3), certainly not presence of SDL1.2  code. I'd be very sad to see SDL1.2 go, especially in the context of 32-bit Atari support. For instance ScummVM, my another Atari project, still supports SDL 1.2 as a first class citizen while having SDL3 code, too. No need to remove anything until there's someone who cares (I do).

ericpareja:
> Carefully with removing that hack -- there are basically three cases to handle (see [libsdl-org/sdl12-compat#331](https://github.com/libsdl-org/sdl12-compat/issues/331)):
> 
>     * original SDL1.2 behaviour
> 
>     * SDL1.2-compat behaviour before 1270 version
> 
>     * SDL1.2-compat behaviour after 1270 version
> 
>     * (and of course SDL2 behaviour)
> 
> 
> So you should build atari800 against all three of them (i.e. you need to install an older/newer sdl1.2 and sdl1.2-compat somewhere and link against that library) to claim this being fixed / improved. Otherwise it would re-introduce the same regressions.

I'll re-write the CAPS LOCK handling patch and submit a separate PR for it without nuking the SDL1.2-compat if possible.

mikrosk:
@ericpareja 

> I'll re-write the CAPS LOCK handling patch and submit a separate PR for it without nuking the SDL1.2-compat if possible.

I didn't mean to discourage you from removing that hack. I just wanted to emphasise the testing process involved. :)

ericpareja:
> @ericpareja
> 
> > I'll re-write the CAPS LOCK handling patch and submit a separate PR for it without nuking the SDL1.2-compat if possible.
> 
> I didn't mean to discourage you from removing that hack. I just wanted to emphasise the testing process involved. :)

I'll submit a separate patch anyway. I have to admit that I forgot that this project is cross-platform, so there are targets that I might be blind about.

PR #268, opened 2026-04-19
Enable NetSIO on Windows via netsiowin.c

pedgarcia:
- Add Win32 UDP implementation (Winsock2, ring-buffer FIFO)
- Changed make scripts to use netsiowin.cpp if the --enable-netsio parameter is set in configure 
- Tested in Fujisan for the past 2 weeks
- Tested with vanilla atari800 built on Windows and executed passing `-netsio` when running it

https://github.com/user-attachments/assets/81fe06ba-db9d-472e-bc67-b56b35a7bdc4


mikrosk:
@pedgarcia please fix the Windows MSYS2 build.

pedgarcia:
Hey @mikrosk - I think the problem is that your CI/CD pipeline doesn't configure with the `./configure --enable-netsio` . How can we change that? (I haven't checked your actions yet)

mikrosk:
@pedgarcia that might be true (feel free to enable it) but when omitted, build shouldn't be failing nevertheless.

pedgarcia:
> @pedgarcia that might be true (feel free to enable it) but when omitted, build shouldn't be failing nevertheless.

Good point. I haven't tried without the flag indeed. I will do that in the next few days. Thanks

pedgarcia:
The last commit makes sure the netsio support is disabled by default for windows as it was before.

mikrosk:
Now I'm even more confused. :) The PR says "enable on windows" but it doesn't build for msys, which is our standard way how to build windows executable?

pedgarcia:
Sorry, maybe I am saying this wrong/confusing. 
- The initial commit made netsio support ON by default on Windows build, which made your PR built to fail. With this change today, I made it OFF by default, so it would pass your build. Users that want netsio support will have to call `configure.sh --enable-netsio <and other parameters>`. 

Maybe we need to update your build in GH Actions to build it with netsio support so it can be trully verified? I wasn't expecting I should do this in this PR (update your GH Actions script), but I can do that if that is what we need?

Note: I just build it locally with and without the parameter and it works as expected.

Let me know if I managed to explain this time :)

mikrosk:
What I still don't understand: you enabled it for windows and it failed on our msys gitlab ci. At the same time, we use msys to build windows executable. So how do _you_ build windows executables?

As I see it the right fix is to investigate why msys is failing _here_ because in the end, this is what release binaries use to produce the final exe.

pedgarcia:
Oh, it is not failing anymore. The changes I made resulting in correct build here. However, the windows CI build GH runs does not exercise the changes I introduced. Maybe windows would need two builds: with and without netsio support?

mikrosk:
@pedgarcia yes, now it is not failing because you have disabled it. So please help me understand, what is this PR about? Providing a windows support which cannot be built? :)

pedgarcia:
**Quick history:** 
- Initial PR enabled NetSIO by default on Windows which mades the  CI failed. 
- Next commit fixed in 4a1e6c2: now disabled-by-default like Unix, but if you build it with **--enable-netsio**  it will add the netsio support

What I think is after this is merged, the CI build could add `./configure --enable-netsio` so the atari800.exe  will be built with full FujiNet users on Windows. Currently your Action runs without any parameter (https://github.com/atari800/atari800/blob/a068f7f85745adf4a63efc877686f47417590f6c/.github/workflows/build.yml#L83).

So if I understand your question right. I did not disabled the netsio support, I made the flag OFF by default. One can still build it with netsio support explicitly passing the flag (--enable-netsio)

mikrosk:
Sorry, it may be just me but I still don't get it. So you changed `A8_OPTION(netsio,yes` to `A8_OPTION(netsio,no`. So netsio is not built, therefore not failing. Now you are saying that all that is needed is to add `--enable-netsio` for MSYS configure and it will be fixed. How? All that `--enable-netsio` does (or it is supposed to do) is to make the `netsio` option `yes` again. Which was exactly the reason why it failed for you.

groessler:
I think CI should build both variants, once with `--enable-netsio` and once with `--disable-netsio`.

pedgarcia:
@mikrosk LOL, I think I finally got what you are talking about - thanks for your patience! 

I think my mistake is that because I managed to build it locally means the netsio support will work here, too. I don't think it will. Looking at the original build, the errors are link errors cause because the windows build uses netsiowin.cpp, and it doesn't export those symbols properly. I will review everything and update the PR.

Let me try to get this done over the weekend...been quite busy with work....

pedgarcia:
@mikrosk The build now should pass not because the netsio enable/disable, but because I added to the netsiowin.cpp the implementation for the missing symbols that the first build link failed (netsio_netstream_*)

Tests:
1) I pushed a change to this PR to force your GH Actions CI to build with netsio support by default - PASS
2) Pushed netsio=off by default as your source is set originally - build also PASS

(*) Unless you guys have a different view, at least for windows I think it is wise to change the CI to build with netsio or like @groessler suggested, have the two builds running so we can detect regressions in the future since the netsiowin implementation is separated from the POSIC netsio (although with some code overlap).

Thanks!

configure.ac
```diff
@@ -978,16 +978,21 @@ A8_OPTION(pbi_bb,yes,
 AM_CONDITIONAL([WANT_PBI_BB], test "$WANT_PBI_BB" = "yes")
 
 if [[ "$SUPPORTS_NETSIO" = "yes" ]]; then
-dnl NetSIO / FujiNet support (only on Linux or MacOS)
-A8_OPTION(netsio,yes,
-       [Support NetSIO Protocol I/O interface (FujiNet) (default=ON)],
+dnl NetSIO / FujiNet support (POSIX: netsio.c; Windows: netsiowin.c). Default OFF so CI builds without it unless --enable-netsio.
```
mikrosk: Didn't you just disable netsio for *all* targets? Also, why? Now that you fixed windows build, is there a reason to have this disabled?

pedgarcia:
This was my mistake - I thought it was off by default... I just reverted it to the original

mikrosk:
Looks good to me, merging. Thanks for addressing all the issues!

pedgarcia:
Awesome, thanks @mikrosk !

PR #269, opened 2026-04-29
netsio fixes for modem, high speed, write_size bug fix. and BASIC Disable fix

mozzwald:
* netsio: make write_size uint16 instead of incorrect uint8
* Fix BASIC disable for OS-initiated XL/XE coldstarts
  * This is needed for Atari side programs that may JMP $E477 (like FujiNet CONFIG does) to restart after mounting disks. It will correctly disable BASIC if the atari800 option is set.
* netsio: fix high speed down to divisor 1
  * event driven netsio_wait_sync instead of sleep delaying
  * netsio gated extra POKEY_siocheck values
* netsio: fix modem serial timing

Side note, I've been working on another project which these fixes were implemented for (in addition to just regular PC use)... a new Android emulator based on atari800 with fujinet-pc: https://github.com/mozzwald/fujinet-go-800
I had made some attempts to get the existing Android emulator updated and building with more modern Android but failed so I started from scratch.

### DIFF
--- a/src/pokey.c
+++ b/src/pokey.c
@@ -174,6 +174,19 @@ static int POKEY_siocheck(void)
 		&& (POKEY_AUDCTL[0] & 0x28) == 0x28;
 }
 
+#ifdef NETSIO
+static int POKEY_serial_byte_delay(void)
+{
+	int divisor = POKEY_AUDF[POKEY_CHAN3];
+
+	if (divisor <= 0)
+		divisor = 1;
+
+	/* Match the existing SIO byte timing formula instead of forcing 19200 baud. */
+	return ((SIO_SERIN_INTERVAL * divisor - 1) / 0x28 + 1);
+}
+#endif /* NETSIO */
+
 #ifndef SOUND_GAIN /* sound gain can be pre-defined in the configure/Makefile */
 #define SOUND_GAIN 4
 #endif
@@ -290,10 +303,17 @@ void POKEY_PutByte(UWORD addr, UBYTE byte)
 
 		/* check if cassette 2-tone mode has been enabled */
 		if ((POKEY_SKCTL & 0x08) == 0x00) {
+#ifdef NETSIO
+			/* Use the active serial divisor for modem/netstream timing. */
+			POKEY_DELAYED_SEROUT_IRQ = POKEY_serial_byte_delay();
+			POKEY_IRQST |= 0x08;
+			POKEY_DELAYED_XMTDONE_IRQ = POKEY_DELAYED_SEROUT_IRQ * 2 - 1;
+#else
 			/* intelligent device */
 			POKEY_DELAYED_SEROUT_IRQ = SIO_SEROUT_INTERVAL;
 			POKEY_IRQST |= 0x08;
 			POKEY_DELAYED_XMTDONE_IRQ = SIO_XMTDONE_INTERVAL;
+#endif /* NETSIO */
 		}
 		else {
 			/* cassette */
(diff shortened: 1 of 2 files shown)


src/gtia.c
```diff
@@ -138,6 +141,44 @@ int hitclr_pos;
...
+void GTIA_ArmBasicDisableBootHold(void)
+{
+	if (!basic_disable_boot_hold_enabled()) {
+		GTIA_ClearBasicDisableBootHold();
+		return;
+	}
+	if (basic_disable_boot_hold_armed && basic_disable_boot_hold_reads_remaining > 0)
+		return;
+	basic_disable_boot_hold_armed = TRUE;
+	basic_disable_boot_hold_reads_remaining = 2;
```
mikrosk: Where this '2' comes from?

src/cpu.c
```diff
@@ -135,7 +136,7 @@ UBYTE CPU_delayed_nmi;
 /* 6502 code fetching */
 #ifdef PC_PTR
 #define GET_PC()            (PC - MEMORY_mem)
-#define SET_PC(newpc)       (PC = MEMORY_mem + (newpc))
+#define SET_PC(newpc)       do { UWORD newpc__ = (newpc); GTIA_CheckForBasicDisableColdstart(newpc__); PC = MEMORY_mem + newpc__; } while (0)
```
mikrosk: Oof, this is a bit heavy-handed. What is the reason for checking this every time a new PC is set?

src/gtia.c
```diff
@@ -138,6 +141,44 @@ int hitclr_pos;
...
+
+void GTIA_ClearBasicDisableBootHold(void)
+{
+	basic_disable_boot_hold_armed = FALSE;
+	basic_disable_boot_hold_reads_remaining = 0;
+}
+
+void GTIA_CheckForBasicDisableColdstart(UWORD pc)
+{
+	if (pc == 0xe477)
```
mikrosk: Doesn't feel right, emulator shouldn't care what address is being executed, only how the (emulated) hardware state is changed.

mikrosk (changes requested):
Overall it is good but I have reservations against the BASIC commit. There must be a cleaner way to implement what are you trying to achieve, listening to PC's value can't be it.

mikrosk:
> I had made some attempts to get the existing Android emulator updated and building with more modern Android but failed so I started from scratch.

What obstacles did you encounter? If it was something hard to update / fix, I'm thinking whether we shouldn't remove Android code altogether, it makes no sense to have dead code sitting in the repo. According to http://users.uoa.gr/~knakos/atari800/android, the last released version was based on Atari800 3.1.0. On the other hand, @kr0tki updated build and readme android files as late as 2020 so I'd expect that about 5 years ago it was still all buildable.

src/gtia.c
mozzwald: If I understood the GTIA console key override, it was checking for START and OPTION which is 2 reads/runs of that function.

src/cpu.c
mozzwald: It was the easiest way I could think of to find out if cold reset was performed from the atari side: watch for a JMP $E477

src/gtia.c
mozzwald: It was the easiest way I could think of to find out if cold reset was performed from the atari side: watch for a JMP $E477

mozzwald:
> What obstacles did you encounter? If it was something hard to update / fix, I'm thinking whether we shouldn't remove Android code altogether, it makes no sense to have dead code sitting in the repo. According to http://users.uoa.gr/~knakos/atari800/android, the last released version was based on Atari800 3.1.0. On the other hand, @kr0tki updated build and readme android files as late as 2020 so I'd expect that about 5 years ago it was still all buildable.

It was over 6 months ago I tried to build the original Android version so I don't remember what issues were there. The biggest problem is that Android has changed so much, the older APK won't install even if built successfully. I was trying to build it with newer Android Studio. This site has a recent build of the original Android emulator, probably built with older toolchains: https://www.zophar.net/android/atari800/colleen.html

I opted to just create a new separate front end in Kotlin with wrappers to atari800/fujinet built with modern Android Studio.

mozzwald:
Regarding the BASIC patch, I will remove it. I could not get atari800 to disable BASIC when FujiNet does a cold reset via JMP $E477 and that was the quickest path to do it. It is pretty invasive. When I get some time, I can revisit it and try to figure out a different way.

mikrosk:
> I could not get atari800 to disable BASIC when FujiNet does a cold reset via JMP $E477

I'm not such an expert on Atari800 internals but why this doesn't work out of the box? I mean, you jump through the same vector which the OS uses for cold start, right. And at some point the OS must check whether the OPTION key has been pressed to enable/disable the built-in BASIC. The GTIA code already handles this flag when reading given byte. So ... why it doesn't work?

mozzwald:
> I'm not such an expert on Atari800 internals but why this doesn't work out of the box? I mean, you jump through the same vector which the OS uses for cold start, right. And at some point the OS must check whether the OPTION key has been pressed to enable/disable the built-in BASIC. The GTIA code already handles this flag when reading given byte. So ... why it doesn't work?

If I remember correctly and from what I could tell, the cold start function took a different path and the JMP was not handled. It's been a bit since I added that and I forget already :D I will revisit it again and open an issue with details.

src/pokey.c
```diff
@@ -511,43 +531,45 @@ void POKEY_Scanline(void)
 
 	if (POKEY_DELAYED_SERIN_IRQ > 0) {
 		if (--POKEY_DELAYED_SERIN_IRQ == 0) {
-			/* Load a byte to SERIN - even when the IRQ is disabled. */
-			POKEY_SERIN = SIO_GetByte();
-			if (POKEY_IRQEN & 0x20) {
-				if (POKEY_IRQST & 0x20) {
-					POKEY_IRQST &= 0xdf;
+#ifdef NETSIO
+			/* Preserve FIFO ordering: don't overwrite SERIN while the previous byte is still pending. */
```
mikrosk, line 543: Why is this important only for netsio and in general, why does it require a special #idef path? (I'm just curious, this is not saying that something is wrong with your code)

mozzwald: Because netsio is network based and it's fifo is managed outside the emulated atari, it could have another byte ready before the atari is ready for it. This delays netsio by a scanline allowing the atari to ack the previous byte. The normal/non-netsio emulated atari path manages it's own state machine and scheduling for sio transfers.

mikrosk:
Looks good to me, if you want to keep the separate commits please force-push the branch with the BASIC commits removed (otherwise I can squash all commits together).

mozzwald:
Removed the BASIC commit and force pushed

